PR #8, opened 2016-02-18
properly define XOPEN_SOURCE CPPFLAGS.

aidancully:
We had -D_XOPEN_SOURCE=600 *and* -D_XOPEN_SOURCE_EXTENDED in CPPFLAGS,
which confuses Illumos's sys/feature-test.h into not defining _XPG5 or
_XPG6. (We had partially worked around this by defining -D_XPG6 on
solaris-type OSes, but this still left _XPG5 undefined, and does not
seem to be the preferred way to define _XPG* feature-test macros.)

### DIFF
--- a/configure.in
+++ b/configure.in
@@ -60,7 +60,7 @@ else
 fi
 SHLDFLAGS="$LDFLAGS"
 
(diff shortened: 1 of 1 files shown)


aidancully:
this change is on the xpgcppflags branch.

PR #10, opened 2016-03-15
add mtev_b<32|64>_<max_de|en>code_len functions and tests.

aidancully:
Also repair b32 encoding so that the encoded data length is always a
multiple of 8 bytes.
Bump library minor version number, because semver.

### DIFF
--- a/src/utils/mtev_b32.h
+++ b/src/utils/mtev_b32.h
@@ -53,6 +53,12 @@
     mtev_b32_decode decodes input until an the entire input is consumed or until an invalid base32 character is encountered.
  */
 API_EXPORT(int) mtev_b32_decode(const char *, size_t, unsigned char *, size_t);
+/*! \fn size_t mtev_b32_decode_len(size_t src_len)
+    \brief Calculate how large a buffer must be to contain a decoded base-32-encoded string of a given length.
+    \param src_len The size (in bytes) of the base-32-encoded string that might be decoded.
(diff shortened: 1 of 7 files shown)


src/utils/mtev_b32.h
```diff
@@ -53,6 +53,12 @@
     mtev_b32_decode decodes input until an the entire input is consumed or until an invalid base32 character is encountered.
  */
 API_EXPORT(int) mtev_b32_decode(const char *, size_t, unsigned char *, size_t);
+/*! \fn size_t mtev_b32_decode_len(size_t src_len)
+    \brief Calculate how large a buffer must be to contain a decoded base-32-encoded string of given length.
```
postwait: "of given length" needs an article... "of a given length"

src/utils/mtev_b64.h
```diff
@@ -53,6 +53,12 @@
     mtev_b64_decode decodes input until an the entire input is consumed or until an invalid base64 character is encountered.
  */
 API_EXPORT(int) mtev_b64_decode(const char *, size_t, unsigned char *, size_t);
+/*! \fn size_t mtev_b64_max_decode_len(size_t src_len)
+    \brief Calculate how large a buffer must be to contain a decoded base-64-encoded string of given length.
```
postwait: "of given length" needs an article... "of a given length"

PR #11, opened 2016-03-18
try to prevent visibility of skiplist placeholder column.

aidancully:
As an implementation detail, each `mtev_skiplist` contains a column
of `mtev_skiplist_node` structures that contain no data. This
detail is intended to be invisible to clients of the skiplist API,
but could be exposed by a few APIs (`mtev_skiplist_find_neighbors`
and friends, `mtev_skiplist_previous`) that manage neighboring
nodes. Modify these routines to stop exposing these placeholder
nodes to clients.

### DIFF
--- a/src/utils/mtev_skiplist.c
+++ b/src/utils/mtev_skiplist.c
@@ -191,14 +191,19 @@ static int mtev_skiplisti_find_compare(mtev_skiplist *sl,
       m=m->next; /* m->next is the match */
       while(m->down) m=m->down; /* proceed to the bottom-most */
       if(ret) *ret = m;
-      if(prev) *prev = m->prev;
+      /* We have to be careful when setting *prev: the first column is a
+       * place-holder element with NULL data that we don't want to expose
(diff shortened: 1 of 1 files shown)


src/utils/mtev_skiplist.c
```diff
@@ -191,14 +191,20 @@ static int mtev_skiplisti_find_compare(mtev_skiplist *sl,
       m=m->next; /* m->next is the match */
       while(m->down) m=m->down; /* proceed to the bottom-most */
       if(ret) *ret = m;
-      if(prev) *prev = m->prev;
+      /* We have to be careful when setting *prev: the first column is a
+       * place-holder element with NULL data that we don't want to expose
+       * to clients. So check that m->prev isn't the first column before
+       * filling in *prev. */
+      if(prev && m->prev && m->prev->prev)
+        *prev = m->prev;
```
postwait: We don't want to skip sething it, instead we want to set it to null.

If (prev)
 *prev = (m->prev && !m->prev->prev) ? NULL : m->prev;

I think

PR #12, opened 2016-03-18
Mtev skiplist node funcs

aidancully:
expose "blessed" interface to remove a node from a skiplist.
add `mtev_skiplist_data` accessor function.

### DIFF
--- a/src/utils/mtev_skiplist.c
+++ b/src/utils/mtev_skiplist.c
@@ -229,6 +229,9 @@ void *mtev_skiplist_previous(mtev_skiplist *sl, mtev_skiplist_node **iter) {
   if(!(*iter)->prev) *iter = NULL;
   return (*iter)?((*iter)->data):NULL;
 }
+void *mtev_skiplist_data(mtev_skiplist_node *m) {
+  return m->data;
+}
 mtev_skiplist_node *mtev_skiplist_insert(mtev_skiplist *sl,
                                          const void *data) {
   if(!sl->compare) return 0;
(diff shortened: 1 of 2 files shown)


src/utils/mtev_skiplist.c
```diff
@@ -37,6 +37,8 @@ static int mtev_skiplisti_find_compare(mtev_skiplist *sl, const void *data,
                                        mtev_skiplist_node **prev,
                                        mtev_skiplist_node **next,
                                        mtev_skiplist_comparator_t comp);
+int mtev_skiplisti_remove(mtev_skiplist *sl, mtev_skiplist_node *m,
```
postwait: Should this be static note that we are hiding it?

src/utils/mtev_skiplist.c
```diff
@@ -227,6 +229,9 @@ void *mtev_skiplist_previous(mtev_skiplist *sl, mtev_skiplist_node **iter) {
   if(!(*iter)->prev) *iter = NULL;
   return (*iter)?((*iter)->data):NULL;
 }
+void *mtev_skiplist_data(mtev_skiplist_node *m) {
+  return m->data;
+}
```
postwait: Does this accessor allow us to make the skiplist node structure opaque?  Just a question,  not a request to do that.

aidancully: I believe it does, but I didn't think we'd want to make that change now.

src/utils/mtev_skiplist.c
aidancully: I wondered about that... Since it was in the header file, it's possible that others were relying on the function... If we accept that possibility, then I think we'd want to bump the version to force downstream clients to re-link against the .so when the function goes away. Keeping it available but out of the header file strongly discourages new clients from using it, but shouldn't cause breakage in old clients that might.

PR #16, opened 2016-04-04
Luamtev

postwait:
Several things all enabling a standalone lua file executer.


src/modules/lua_general.c
```diff
@@ -105,6 +115,7 @@ lua_general_resume(mtev_lua_resume_info_t *ri, int nargs) {
         if(lua_isstring(ri->coro_state, base-1)) {
           err = lua_tostring(ri->coro_state, base-1);
           mtevL(nlerr, "err -> %s\n", err);
+          tragic_failure(ri->lmc->self);
```
aidancully: I think the `tragic_failure` belongs outside this `if` block? probably just above `rv = -1`, below?

PR #22, opened 2016-04-18
added mtev_confstr_parse_time_gm routine.

aidancully:
parses rfc3339 datespecs.

also, for consistency, changed existing mtev_confstr to return
MTEV_CONFSTR_PARSE result codes.

### DIFF
--- a/src/utils/mtev_confstr.c
+++ b/src/utils/mtev_confstr.c
@@ -1,4 +1,5 @@
 #include "mtev_confstr.h"
+#include <time.h>
 #include <ctype.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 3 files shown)


src/utils/mtev_confstr.c
```diff
@@ -122,13 +123,13 @@ int
 mtev_confstr_parse_boolean(const char *input, mtev_boolean *output) {
   if(!strcasecmp(input, "yes") || !strcasecmp(input, "true") || !strcasecmp(input, "on")) {
```
rileyberton: Maybe also support "1"?

src/utils/mtev_confstr.c
```diff
@@ -122,13 +123,13 @@ int
...
     *output = mtev_true;
-    return 1;
+    return MTEV_CONFSTR_PARSE_SUCCESS;
   }
   if(!strcasecmp(input, "no") || !strcasecmp(input, "false") || !strcasecmp(input, "off")) {
```
rileyberton: Maybe also support "0"?

src/utils/mtev_confstr.c
aidancully: Maybe? but I'd put that in a different PR. For what it's worth, those strings were mostly copied from [reconnoiter](https://github.com/circonus-labs/reconnoiter/blob/master/src/modules/httptrap.c#L111).

rileyberton:
Looks good to me.

src/utils/mtev_confstr.c
postwait: Reconnoiter and thus libmtev use on/true and off/ false. If add 1/0, I
agree it should be a separate and sweeping commit.

PR #23, opened 2016-04-21
portability fixes.

aidancully:


src/eventer/eventer_SSL_fd_opset.c
```diff
@@ -671,8 +671,10 @@ eventer_ssl_ctx_new(eventer_ssl_orientation_t type,
     populate_finfo(&ctx->ssl_ctx_cn->key_finfo, key);
     populate_finfo(&ctx->ssl_ctx_cn->ca_finfo, ca);
     ctx->ssl_ctx = NULL;
+    if(0)
+      ;
 #if defined(SSL_TXT_SSLV3) && defined(HAVE_SSLV3_SERVER) && defined(HAVE_SSLV3_CLIENT)
```
rileyberton: I don't get this one.

aidancully: If the first `#if defined() etc.` block compiled out, and a later one compiled in, then that later one (which begins with an `else if`) would generate a syntax error, because there would be no leading `if`. So, we create an `if` clause for all blocks to hang off of. Gotta love the pre-processor.

PR #25, opened 2016-04-29
Implemented mtev_conf_check

JonasKunze:
This allows you to check the existens of configs in a clean way. An example could look like following:

  struct mtev_conf_description_t descs[] = { {
  CONFIG_FQ_HOST, MTEV_CONF_TYPE_STRING,
      "Hostname of the fq broker data should be received from" },

  { CONFIG_FQ_PORT, MTEV_CONF_TYPE_INT, "Port number of the fq broker" },

  { CONFIG_FQ_USER, MTEV_CONF_TYPE_STRING,
      "User name used to connect to the fq broker" },

  { CONFIG_FQ_PASS, MTEV_CONF_TYPE_STRING,
      "Password used to connect to the fq broker" } };

  if (mtev_conf_load_desc(config_file, descs,
      sizeof(descs) / sizeof(struct mtev_conf_description_t)) != 0) {
    mtevL(mtev_error, "Cannot load config: '%s'\n", config_file);
    exit(2);
  }

If a config path isn't found this will print a message explaining which path is missing and what's the semantics of the corresponding value.

---
This change is [Reviewable](https://reviewable.io/reviews/circonus-labs/libmtev/25)

### DIFF
--- a/src/mtev_conf.c
+++ b/src/mtev_conf.c
@@ -1063,7 +1063,72 @@ mtev_conf_load(const char *path) {
   XML2LOG(xml_debug);
   return rv;
 }
+static int mtev_conf_check_value(mtev_conf_description_t* description) {
+  int rv = 0;
+  switch (description->type) {
+  case MTEV_CONF_TYPE_BOOLEAN:
+    rv = mtev_conf_get_boolean(description->section, description->path,
+        &description->value.val_bool);
+    break;
+  case MTEV_CONF_TYPE_INT:
+    rv = mtev_conf_get_int(description->section, description->path,
+        &description->value.val_int);
+    break;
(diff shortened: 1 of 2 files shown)


src/mtev_conf.c
```diff
@@ -1064,6 +1064,67 @@ mtev_conf_load(const char *path) {
...
+  default:
+    rv = -1;
+  }
+  return rv;
+}
+
+int
+mtev_conf_check(mtev_conf_description_t* descriptions, int descriptions_cnt) {
+  for (int i = 0; i != descriptions_cnt; i++) {
+    if (!mtev_conf_check_type(descriptions[i].section, descriptions[i].path,
```
rileyberton: Can we check for explicit return value here instead of `! stuff`?`

PR #28, opened 2016-05-02
Implemented mtev_conf_section_to_xpath

JonasKunze:
This is used by mtev_conf_check to print nicer error messages

### DIFF
--- a/src/mtev_conf.c
+++ b/src/mtev_conf.c
@@ -54,6 +54,7 @@
 #include "mtev_watchdog.h"
 #include "mtev_security.h"
 #include "mtev_hooks.h"
+#include "mtev_str.h"
 
 MTEV_HOOK_IMPL(mtev_conf_delete_section,
                 (const char *root, const char *path,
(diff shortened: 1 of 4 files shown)


src/mtev_conf.c
```diff
@@ -1100,21 +1100,50 @@ static int mtev_conf_check_value(mtev_conf_description_t* description) {
...
 }
 
+char*
+mtev_conf_section_to_xpath(mtev_conf_section_t* section) {
+  if (section == NULL) {
+    return NULL;
+  }
+
+  const int max_len = 512;
+  char* string = alloca(max_len);
```
postwait: I think 512 is a tad shy... you should run two loops through, the first time to determine the size.

src/mtev_conf.c
```diff
@@ -1100,21 +1100,50 @@ static int mtev_conf_check_value(mtev_conf_description_t* description) {
...
+  char* string_ptr = string + max_len;
+
+  xmlNode* node = (xmlNode*) section;
+  while (node && node->name) {
+    int len = strlen(node->name);
+    string_ptr -= len+1;
+    if(string_ptr < string) {
+      break;
+    }
+    memcpy(string_ptr+1, node->name, len);
```
postwait: /foo/bar/baz isn't an unambiguous XPath name.  Think of documents with multiple bar and multiple baz nodes.

PR #29, opened 2016-05-03
fixed some gcc warning messages

JonasKunze:


src/mtev_cluster.c
```diff
@@ -170,9 +171,6 @@ mtev_cluster_update_config(mtev_cluster_t *cluster, mtev_boolean create) {
   mtev_conf_request_write();
 
   return 1;
- bail:
-  if(create && parent) xmlFree(parent);
-  return 0;
```
JonasKunze: This wasn't used anywhere. I'm not entirely sure if it's really not needed any more or just accidentally not used.

PR #33, opened 2016-05-04
checking if an <appname> tag exists in the config

JonasKunze:


src/mtev_main.c
```diff
@@ -164,6 +164,16 @@ mtev_main(const char *appname,
     exit(-1);
   }
 
+  int cnt;
+  char* root_section_path = alloca(strlen(appname)+2);
+  sprintf(root_section_path, "/%s", appname);
+  mtev_conf_get_sections(NULL, root_section_path, &cnt);
+
+  if(cnt==0) {
+    fprintf(stderr, "The config must contain a <%s> tag\n", appname);
```
postwait: Change this to "The config must have <%s> as its root node.\n"

PR #34, opened 2016-05-05
Bugfix: segfault and memory leak when using mtev_cluster

JonasKunze:


postwait:
This was fixed as a part of PR #35

PR #41, opened 2016-05-18
ernie-ping

JonasKunze:
Implemented a very basic algorithm that determines the oldest node in a cluster:
1) Every node knows the boot times of all other nodes in the cluster
2) If two nodes have the identic boot time, they'll set the `tv_usec` to `random()%1E6`
3) If the `last_contact` time is of a node is older than `2*period`, it's boot time is set to a large value in the future and the new oldest node is determined

### DIFF
--- a/src/mtev_cluster.c
+++ b/src/mtev_cluster.c
@@ -48,18 +48,31 @@ static struct timeval my_boot_time;
 static mtev_boolean have_clusters;
 static mtev_hash_table global_clusters;
 
+static const struct timeval boot_time_of_dead_node = { (1UL
+    << (CHAR_BIT * sizeof(boot_time_of_dead_node.tv_sec) - 1)) - 1, (1UL
+    << (CHAR_BIT * sizeof(boot_time_of_dead_node.tv_usec) - 1)) - 1 };
+
 /* All allocated with mtev_memory_safe commands */
 struct mtev_cluster_t {
   char *name;
   unsigned short port;
   int period;
+  int timeout;
+  int maturity;
   char *key;
   int64_t config_seq;
   int node_cnt;
   mtev_cluster_node_t *nodes;
   mtev_cht_t **cht;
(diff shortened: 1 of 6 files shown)


src/mtev_cluster.c
```diff
@@ -265,6 +359,8 @@ mtev_cluster_update_config(mtev_cluster_t *cluster, mtev_boolean create) {
     xmlSetProp(parent, (xmlChar *)"port", (xmlChar *)port);
     snprintf(period, sizeof(period), "%d", cluster->period);
     xmlSetProp(parent, (xmlChar *)"period", (xmlChar *)period);
+    snprintf(timeout_ms, sizeof(timeout_ms), "%d", cluster->timeout_ms);
+    xmlSetProp(parent, (xmlChar *)"timeout_ms", (xmlChar *)period);
```
postwait: All times in libmtev (like period, and timeout) are millisecond, having just this one with a _ms ending breaks that consistency.  change to "timeout."

Also, I think that a "maturity" variable would be good, it can default to "timeout" but could be specified separately.  Only nodes that have seen last_context within "timeout" and have a boot time older than "maturity"

src/mtev_cluster.c
```diff
@@ -359,6 +459,27 @@ mtev_cluster_update_internal(mtev_conf_section_t cluster,
...
+  }
+  timeout = strtoll(bufstr, &endptr, 10);
+  if(*endptr || timeout < period) {
+    mtevL(mtev_error, "Cluster '%s' timeout invalid.\n", name);
+    goto bail;
+  }
+
+  if(!mtev_conf_get_stringbuf(cluster, "@maturity", bufstr, sizeof(bufstr)) ||
+     bufstr[0] == '\0') {
+     strlcpy(bufstr, "5000", sizeof(bufstr));
```
postwait: this should be the timeout... not the default timeout.

PR #44, opened 2016-05-20
Support stack trace print out on errors.

postwait:


src/modules/lua.c
```diff
@@ -250,6 +250,18 @@ register_console_lua_commands() {
     NCSCMD("lua", mtev_console_show_lua, NULL, NULL, NULL));
 }
 
+int
+mtev_lua_traceback(lua_State *L) {
+  if (!lua_isstring(L, 1)) {
```
aidancully: shouldn't the index of most of these be `-1`?

src/modules/lua_general.c
```diff
@@ -156,8 +158,9 @@ lua_general_handler(mtev_dso_generic_t *self) {
   if(rv) {
     int i;
     mtevL(nlerr, "lua: require %s failed\n", conf->module);
+    mtev_lua_traceback(L);
     i = lua_gettop(L);
-    if(i>0) {
+    if(i>=0) {
```
aidancully: with this change, we might call `lua_isstring(L, 0)`, isn't that disallowed?

src/modules/lua_web.c
```diff
@@ -212,8 +214,9 @@ lua_web_handler(mtev_http_rest_closure_t *restc,
   if(rv) {
     int i;
     mtevL(mtev_error, "lua: require %s failed\n", restc->closure);
+    mtev_lua_traceback(L);
     i = lua_gettop(L);
-    if(i>0) {
+    if(i>=0) {
```
aidancully: with this change, we might call lua_isstring(L, 0), isn't that disallowed?

PR #53, opened 2016-06-02
Undef "user2" If term.h Is Included

pamaddox:
term.h #defines "user2", which conflicts with some libraries (notably -
libxml and the libxml/xterm.h include file)... undefine it if we had to
include term.h.


aidancully:
LGTM

PR #54, opened 2016-06-02
Move Including #term.h Into mtev_console_telnet C File

pamaddox:
Rather than doing an #undef for user2 (there are tons of bad defines in
term.h), just move term.h include into the C file itself.


aidancully:
LGTM - in fact, even better than PR #53!

PR #55, opened 2016-06-02
Move curses.h Include To mtev_console_telnet.c

pamaddox:
Take curses.h out of mtev_console_telnet.h.... were getting conflicts
with the macro "move".


aidancully:
LGTM

PR #63, opened 2016-06-17
a timeout callback may be passed to socket:read used by HttpClient

JonasKunze:


src/modules/lua_mtev.c
```diff
@@ -1155,18 +1204,40 @@ mtev_lua_socket_read(lua_State *L) {
...
+    eventer_remove_fd(e->fd);
+    e->callback = mtev_lua_socket_read_complete;
+    e->mask = mask | EVENTER_EXCEPTION;
+    eventer_add(e);
+
+    if (lua_gettop(L) == 5 && lua_isfunction(L, 5)) {
+      int timeout;
+      timeout = 10;
+      if(lua_isnumber(L, 4)) {
+        timeout = lua_tointeger(L, 4);
```
postwait: We should allow a double here representing sub-second values.

There are a lot of times when a read might want a 100ms timeout. "0.1"

src/modules/lua_mtev.c
```diff
@@ -1155,18 +1204,40 @@ mtev_lua_socket_read(lua_State *L) {
...
+      gettimeofday(&timeout_eventer->whence, NULL);
+      timeout_eventer->whence.tv_sec += timeout;
+      timeout_eventer->callback = on_timeout;
+      timeout_eventer->closure = cb_ref;
+      eventer_add_timed(timeout_eventer);
+
+      cl->timeout_event = timeout_eventer;
+    }
+
+    return mtev_lua_yield(ci, 0);}
```
postwait: Move } to next line.

src/modules/lua_mtev.c
```diff
@@ -1094,6 +1107,42 @@ mtev_lua_socket_read_complete(eventer_t e, int mask, void *vcl,
...
+
+  cl = cb_ref->timed_out_eventer->closure;
+  ci = mtev_lua_get_resume_info(L);
+  assert(ci);
+
+  // remove the original read event
+  eventer_remove_fd(cb_ref->timed_out_eventer->fd);
+  mtev_lua_deregister_event(ci, cb_ref->timed_out_eventer, 0);
+
+  // resume the original e:read() call and return nil
```
postwait: This comment is misleading.  on_timeout will likely be used for other timeout socket ops... 

The subsequent four lines of code are part of the removal above... make them a contiguous block of 6 lines.

put this comment right before the lua_pushnil(L) and say something like...

// return into lua wherever we might have been blocking

src/modules/lua_mtev.h
```diff
@@ -99,6 +105,7 @@ struct nl_slcl {
   size_t write_goal;
   eventer_t *eptr;
   eventer_t pending_event;
+  eventer_t timeout_event;
```
postwait: How does this get cleaned up if the whole coro is cancelled?

src/modules/lua_mtev.c
```diff
@@ -1094,6 +1107,41 @@ mtev_lua_socket_read_complete(eventer_t e, int mask, void *vcl,
...
+  L = cb_ref->L;
+
+  // run the timeout callback
+  lua_rawgeti( L, LUA_REGISTRYINDEX, cb_ref->callback_reference );
+  lua_call(L, 0, 0);
+
+  cl = cb_ref->timed_out_eventer->closure;
+  ci = mtev_lua_get_resume_info(L);
+  assert(ci);
+
```
postwait: deregister e from ci here.

src/modules/lua_mtev.c
```diff
@@ -1064,6 +1070,13 @@ mtev_lua_socket_read_complete(eventer_t e, int mask, void *vcl,
   int len;
   int args = 0;
 
+  if(cl->timeout_event) {
+    eventer_remove_timed(cl->timeout_event);
+    lua_timeout_callback_ref_free(cl->timeout_event->closure);
```
postwait: deregister the event with lua here. (you need the mtev_lua_resume_info_t to do that)

src/modules/lua_mtev.c
```diff
@@ -1155,18 +1203,47 @@ mtev_lua_socket_read(lua_State *L) {
...
+      cb_ref->timed_out_eventer = e;
+
+      eventer_t timeout_eventer = eventer_alloc();
+      timeout_eventer->mask = EVENTER_TIMER;
+      gettimeofday(&timeout_eventer->whence, NULL);
+      timeout_eventer->whence.tv_sec += timeout_s;
+      timeout_eventer->whence.tv_usec += timeout_us;
+      timeout_eventer->callback = on_timeout;
+      timeout_eventer->closure = cb_ref;
+      eventer_add_timed(timeout_eventer);
```
postwait: register the event in the ci before the add_timed.

postwait:
Can you put this through another round of testing so you are confident it works the way you anticipated?  The code looks good to me.

JonasKunze:
this has already been tested via the unit tests of beaker!

JonasKunze:
Comment from Theo:
the timeout you set is on every read… instead of the request.
which is unintuitive.
I think you want to set the timeout from the start of the do_request… and adjust it for future reads so that you aim at the same timeout deadline.

PR #65, opened 2016-06-17
Support C-loaded module names like namespace:foo.

postwait:
* namespace:foo will just load "foo", but register the whole name.
  * names should be returned by hash key, not by internal name


src/mtev_dso.c
```diff
@@ -102,7 +102,7 @@ mtev_dso_list(mtev_hash_table *t, const char ***f) {
   while(mtev_hash_next(t, &iter, (const char **)&name, &klen,
                        &vhdr)) {
     mtev_image_t *hdr = (mtev_image_t *)vhdr;
```
aidancully: delete this line - nothing uses `vhdr` or `hdr` any more

PR #66, opened 2016-06-20
Thread Safety Fixes To Reverse Socket Code

pamaddox:
Added reference counts to reverse socket code to avoid use-after-free
issues. Changed a couple of locks to avoid multiple threads accessing
the same code.

### DIFF
--- a/src/mtev_reverse_socket.c
+++ b/src/mtev_reverse_socket.c
@@ -127,14 +127,12 @@ static void reverse_frame_free(void *vrf) {
 }
 
 typedef struct {
-  char *id;
   int up;
   struct timeval create_time;
   uint64_t in_bytes;
   uint64_t in_frames;
   uint64_t out_bytes;
   uint64_t out_frames;
-  pthread_mutex_t lock;
   eventer_t e;
   reverse_frame_t *outgoing;
   reverse_frame_t *outgoing_tail;
@@ -180,6 +178,14 @@ typedef struct {
 
   /* If we have one of these, we need to refresh timeouts */
   mtev_connection_ctx_t *nctx;
+
+} reverse_socket_data_t;
+
+typedef struct {
+  reverse_socket_data_t data;
+  char *id;
+  pthread_mutex_t lock;
+  mtev_atomic32_t refcnt;
 } reverse_socket_t;
 
 typedef struct {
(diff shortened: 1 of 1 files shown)


src/mtev_reverse_socket.c
```diff
@@ -287,71 +308,75 @@ command_out(reverse_socket_t *rc, uint16_t id, const char *command) {
...
+  if(rc->data.channels[i].pair[0] >= 0) {
     mtevL(nldeb, "mtev_reverse_socket_channel_shutdown(%s, %d)\n", rc->id, i);
+    need_deref = 1;
+  }
   pthread_mutex_lock(&rc->lock);
-  if(rc->channels[i].pair[0] >= 0) {
-    int fd = rc->channels[i].pair[0];
+  if(rc->data.channels[i].pair[0] >= 0) {
+    int fd = rc->data.channels[i].pair[0];
     eventer_t ce = eventer_find_fd(fd);
```
postwait: This ce maskes the top-level ce.   I do'nt think this code does what we think it does.

src/mtev_reverse_socket.c
```diff
@@ -287,71 +308,75 @@ command_out(reverse_socket_t *rc, uint16_t id, const char *command) {
...
-    rc->channels[i].pair[0] = -1;
+    rc->data.channels[i].pair[0] = -1;
     if(!ce) close(fd);
   }
   pthread_mutex_unlock(&rc->lock);
 
   if(ce) eventer_trigger(ce, EVENTER_EXCEPTION);
```
postwait: This ce is always null do to variable scope shadowing.

aidancully:
LGTM

src/mtev_reverse_socket.c
```diff
@@ -387,6 +412,7 @@ mtev_reverse_socket_channel_handler(eventer_t e, int mask, void *closure,
     eventer_remove_fd(e->fd);
     CHANNEL.pair[0] = CHANNEL.pair[1] = -1;
     mtev_reverse_socket_channel_shutdown(cct->parent, cct->channel_id, e);
+    mtev_reverse_socket_deref(cct->parent);
```
postwait: So the channl_shutdown right before this can possibly deref as well... Is a double deref in that case intended?

src/mtev_reverse_socket.c
aidancully: test line note

src/mtev_reverse_socket.c
rileyberton: test line note.

src/mtev_reverse_socket.c
```diff
@@ -180,26 +178,21 @@ typedef struct {
 
   /* If we have one of these, we need to refresh timeouts */
   mtev_connection_ctx_t *nctx;
+
+} reverse_socket_data_t;
+
```
rileyberton: test line note.

src/mtev_reverse_socket.c
pamaddox: We are setting channel pairs to -1 right before the call..... we only deref in the function if those are not -1, so we won't deref in the function here.

src/mtev_reverse_socket.c
pamaddox: Yeah, that seems broken. Not directly related to the PR at hand, but it's something worth looking into. I'll investigate it.

PR #67, opened 2016-06-21
Fix Shadowed Argument Bug

pamaddox:
We were declaring a local version of a variable that was shadowing
another variable, which was causing us to never look at the new value.
Fix this to work properly and change the reverse socket deref stuff to
reflect this fix.


postwait:
LGTM

PR #68, opened 2016-06-21
Fix A Busted Reference Count In Eventer SSL Code

pamaddox:
We were double-decrementing a refcnt, which could lead to undocumented
behavior. Fix it to decrement properly.


aidancully:
LGTM

PR #70, opened 2016-06-21
Disable ECDH

pamaddox:
Disable ECDH stuff temporarily until we can find the source of an SSL
bug we're seeing.


pamaddox:
Eh, think that we want to actually fix this rather than just work around. Gonna close this and work on a fix.

PR #73, opened 2016-06-22
Better Error Handling Before Aborts

pamaddox:
When we abort, rather than just aborting, log what happens via the new
MTEV_LOG_AND_ABORT() function. This will set the log to synchronous, log
the message, then abort. This will avoid the issue where we queue
something to be written to a log but exit before it flushes.

### DIFF
--- a/src/eventer/eventer_epoll_impl.c
+++ b/src/eventer/eventer_epoll_impl.c
@@ -179,13 +179,10 @@ static eventer_t eventer_epoll_impl_remove(eventer_t e) {
       removed = e;
       master_fds[e->fd].e = NULL;
       if(epoll_ctl(spec->epoll_fd, EPOLL_CTL_DEL, e->fd, &_ev) != 0) {
+        mtevL(mtev_error, "epoll_ctl(%d, EPOLL_CTL_DEL, %d) -> %s\n",
+              spec->epoll_fd, e->fd, strerror(errno));
         if(errno != ENOENT) {
-          mtevFatal(mtev_error, "epoll_ctl(%d, EPOLL_CTL_DEL, %d) -> %s\n",
-                spec->epoll_fd, e->fd, strerror(errno));
-        }
-        else {
-          mtevL(mtev_error, "epoll_ctl(%d, EPOLL_CTL_DEL, %d) -> %s\n",
-                spec->epoll_fd, e->fd, strerror(errno));
+          mtevFatal(mtev_error, "errno != ENOENT: %d (%s)\n", errno, strerror(errno));
         }
       }
     }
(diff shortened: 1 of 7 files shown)


src/utils/mtev_log.h
```diff
@@ -171,6 +171,11 @@ API_EXPORT(int)
 #define SETUP_LOG(a, b) do { if(!a##_log) a##_log = mtev_log_stream_find(#a); \
                              if(!a##_log) { b; } } while(0)
 
+#define MTEV_LOG_AND_ABORT(log,args...) \
+  mtev_log_go_synch(); \
+  mtevL(log, args); \
+  abort();
+
```
aidancully: this macro should be wrapped in a `do {...} while(0)`, with no trailing semicolon... consider:
```
if(foo)
  MTEV_LOG_AND_ABORT(mtev_error, "blah");
```
which, with the current form, would only put `mtev_log_go_synch();` inside the `if` condition, and execute the next two statements unconditionally.

pamaddox: Good call. Fixed.

aidancully:
LGTM

src/eventer/eventer_epoll_impl.c
```diff
@@ -177,9 +179,14 @@ static eventer_t eventer_epoll_impl_remove(eventer_t e) {
...
-              spec->epoll_fd, e->fd, strerror(errno));
-        if(errno != ENOENT) abort();
+        if(errno != ENOENT) {
+          mtevFatal(mtev_error, "epoll_ctl(%d, EPOLL_CTL_DEL, %d) -> %s\n",
+                spec->epoll_fd, e->fd, strerror(errno));
+        }
+        else {
+          mtevL(mtev_error, "epoll_ctl(%d, EPOLL_CTL_DEL, %d) -> %s\n",
+                spec->epoll_fd, e->fd, strerror(errno));
+        }
```
aidancully: Stylistically, the Cut-n-Paste log message bothers me... Not a big deal, but what if you left the old `mtevL()` call, then had `mtevFatal(mtev_error, "errno != ENOENT: %d (%s)\n", errno, strerror(errno));` in the abort case?

src/eventer/eventer_epoll_impl.c
```diff
@@ -230,9 +237,14 @@ static eventer_t eventer_epoll_impl_remove_fd(int fd) {
...
-            spec->epoll_fd, fd, strerror(errno));
-      if(errno != ENOENT) abort();
+      if(errno != ENOENT) {
+        mtevFatal(mtev_error, "epoll_ctl(%d, EPOLL_CTL_DEL, %d) -> %s\n",
+              spec->epoll_fd, fd, strerror(errno));
+      }
+      else {
+        mtevL(mtev_error, "epoll_ctl(%d, EPOLL_CTL_DEL, %d) -> %s\n",
+              spec->epoll_fd, fd, strerror(errno));
+      }
```
aidancully: see note above about cut-n-paste for the log message

pamaddox: changed

aidancully:
LGTM

src/eventer/eventer_epoll_impl.c
pamaddox: changed

PR #75, opened 2016-06-22
Websocket support using wslay

rileyberton:
http://wslay.sourceforge.net/

A bunch of unrelated changes got pulled in because of an accidental rebase from master.

This add support for "ws://" and "wss://" and ability to write apps that communicate via websockets.  This is layered into the normal `mtev_http_rest` API and so will degrade to a normal HTTP session if something goes wrong with the websocket.  This also means that it support all the rest authentication that a normal mtev rest app supports.

Currently I am only dispatching to user defined handlers when we receive a complete websocket message.  This might hamstring people who need finer grained control over websocket communication but we can always expose this level of control in a later release.


rileyberton:
Closing this to create another branch that cleans up the accidental rebase

PR #76, opened 2016-06-22
Websocket support using wslay

rileyberton:
http://wslay.sourceforge.net/

This add support for "ws://" and "wss://" and ability to write apps that communicate via websockets. This is layered into the normal `mtev_http_rest` API and so will degrade to a normal HTTP session if something goes wrong with the websocket. This also means that it support all the rest authentication that a normal mtev rest app supports.

Currently I am only dispatching to user defined handlers when we receive a complete websocket message. This might hamstring people who need finer grained control over websocket communication but we can always expose this level of control in a later release.


src/examples/fq-router.c
```diff
@@ -74,7 +76,7 @@ parse_cli_args(int argc, char * const *argv) {
   }
 }
 static mtev_hook_return_t
-on_msg_received(void *closure, fq_client client, int connection_id, fq_msg *message) {
+on_msg_received(void *closure, fq_client client, int connection_id, fq_msg *message, void * a, size_t b) {
```
rileyberton: This example program had lagged behind ABI changes so wouldn't compile cleanly.

src/mtev_http.c
```diff
@@ -1184,6 +1226,162 @@ mtev_http_session_req_consume(mtev_http_session_ctx *ctx,
...
+#define WS_ACCEPT_KEY_GUID "258EAFA5-E914-47DA-95CA-C5AB0DC85B11"
+
+void
+mtev_http_create_websocket_accept_key(char *dest, size_t dest_len, const char *client_key)
+{
+  SHA_CTX ctx;
+  unsigned char sha1[SHA_DIGEST_LENGTH], key_src[60];
+
+  memcpy(key_src, client_key, 24);
+  memcpy(key_src+24, WS_ACCEPT_KEY_GUID, 36);
```
aidancully: can these magic constants (`24`, `36`) be `#define`s? The `36` should presumably be `UUID_STR_LEN`, not off-hand sure where `24` comes from...

src/mtev_http.c
```diff
@@ -1184,6 +1226,162 @@ mtev_http_session_req_consume(mtev_http_session_ctx *ctx,
...
+
+  SHA1_Init(&ctx);
+  SHA1_Update(&ctx, (const void *)key_src, (unsigned long)sizeof(key_src));
+  SHA1_Final(sha1, &ctx);
+
+  mtev_b64_encode(sha1, SHA_DIGEST_LENGTH, dest, dest_len);
+  dest[mtev_b64_encode_len(SHA_DIGEST_LENGTH)] = '\0';
```
aidancully: should we have an `mtevAssert` that `dest_len >= mtev_64_encode_len(SHA_DIGEST_LENGTH)`?

src/mtev_http.c
```diff
@@ -1184,6 +1226,162 @@ mtev_http_session_req_consume(mtev_http_session_ctx *ctx,
...
+    return ctx->is_websocket;
+  }
+
+  mtev_http_create_websocket_accept_key(accept_key, sizeof(accept_key), sec_ws_key);
+  
+  /* now we upgrade their socket */
+  mtev_http_response_header_set(ctx, "Upgrade", "websocket");
+  mtev_http_response_header_set(ctx, "Connection", "Upgrade");
+  mtev_http_response_header_set(ctx, "Sec-WebSocket-Accept", accept_key);
+  mtev_http_response_header_set(ctx, "Sec-WebSocket-Protocol", protocol);
```
aidancully: We haven't verified that `protocol` was set, by this point, maybe it's NULL?

src/mtev_http.c
```diff
@@ -1280,8 +1529,9 @@ mtev_http_session_drive(eventer_t e, int origmask, void *closure,
 }
 
 mtev_http_session_ctx *
-mtev_http_session_ctx_new(mtev_http_dispatch_func f, void *c, eventer_t e,
-                          acceptor_closure_t *ac) {
+mtev_http_session_ctx_new(mtev_http_dispatch_func f, mtev_http_websocket_dispatch_func wf,
+                          void *c, eventer_t e, acceptor_closure_t *ac) 
+{
```
aidancully: nothing outside libmtev calls this function?

src/mtev_rest.c
```diff
@@ -196,16 +200,27 @@ mtev_http_get_handler(mtev_http_rest_closure_t *restc) {
     }
     eob--;
   }
+
+  /* no base, give up */
   if(!cont) return NULL;
+
+  const char *protocol = NULL;
+  mtev_hash_table *headers = mtev_http_request_headers_table(req);
+
```
aidancully: stylistically, I think we've been putting local variable definitions in C code directly after open braces.

src/mtev_rest.c
```diff
@@ -274,18 +320,67 @@ mtev_http_rest_disclose_endpoints(const char *base, const char *expr) {
...
 int
 mtev_http_rest_register_auth_closure(const char *method, const char *base,
-                             const char *expr, rest_request_handler f,
-                             rest_authorize_func_t auth, void *closure) {
+                                     const char *expr, rest_request_handler f,
+                                     rest_websocket_message_handler wf, 
+                                     const char *websocket_protocol, 
+                                     rest_authorize_func_t auth, void *closure) 
+{
   void *vcont;
```
aidancully: nothing outside libmtev calls this function?

src/mtev_http.c
rileyberton: This is a good point.. adding a wrapper function to keep ABI compat

src/mtev_rest.c
rileyberton: Not in anything I have grepped.. external users might but I doubt this.

postwait:
+1

src/examples/websocket_server.conf
```diff
@@ -0,0 +1,67 @@
...
+  <watchdog glider="/opt/local/bin/bt"
+            tracedir="/var/cores/example"/>
+  -->
+  <eventer>
+    <config>
+      <concurrency>4</concurrency>
+      <default_queue_threads>10</default_queue_threads>
+      <default_ca_chain>/etc/default-ca-chain.crt</default_ca_chain>
+      <ssl_dhparam512_file>/var/run/example/dhparam512.txt</ssl_dhparam512_file>
+      <ssl_dhparam1024_file>/var/run/example/dhparam1024.txt</ssl_dhparam1024_file>
```
esproul: FYI, after https://github.com/circonus-labs/libmtev/pull/84 we will no longer support 512-bit DH params and will support 2048-bit ones, so if you want, maybe s/512/2048/ here.  Not a big deal though.

PR #77, opened 2016-06-23
a timeout callback may be passed to socket:read used by HttpClient

JonasKunze:
e.timeout now as double
registering timeout events so that they are cleaned up when a coro is canceled
HttpClient timeout counts for a request instead of a read call

### DIFF
--- a/src/modules/lua-support/mtev/HttpClient.lua
+++ b/src/modules/lua-support/mtev/HttpClient.lua
@@ -79,6 +79,13 @@ function HttpClient:connect(target, port, ssl, ssl_host, ssl_layer)
                                      ssl_host, ssl_layer)
 end
 
+function HttpClient:close()
+  if self.e ~= nil then
+    self.e:close()
+    self.e = nil
+  end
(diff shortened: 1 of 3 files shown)


src/modules/lua-support/mtev/HttpClient.lua
```diff
@@ -107,11 +114,41 @@ function HttpClient:do_request(method, uri, _headers, payload, http_version)
       sstr = sstr .. payload
     end
     self.e:write(sstr)
+
+    if self.on_timeout ~= nil then
+      self.request_started_s, self.request_started_us = mtev.gettimeofday()
+    end
```
postwait: I would set these unconditionally.

Now, if someone does a do_request and sets a timeout in a callback, you'll attempt to put a callback in place with no request start time.  Seems easier to just always record the request start time instead of preventing that (possibly) legitimate use.

JonasKunze: good point!

PR #78, opened 2016-06-23
Replace "assert" Calls With "mtevAssert"

pamaddox:
Add a custom call, "mtevAssert", that will replace normal assert calls.
This will log the error to mtevError, flush the log, then abort.

In addition, this makes it much easier if we wish to add other prints or
information to the logging when we need to assert.

### DIFF
--- a/src/eventer/eventer_ports_impl.c
+++ b/src/eventer/eventer_ports_impl.c
@@ -116,7 +116,7 @@ static void alter_fd_dissociate(eventer_t e, int mask, struct ports_spec *spec) 
   if (ret == -1) {
     if(s_errno == ENOENT) return; /* Fine */
     if(s_errno == EBADFD) return; /* Fine */
-    mtevL(mtev_error,
+    mtevFatal(mtev_error,
           "eventer port_dissociate failed(%d-%d): %d/%s\n", e->fd, spec->port_fd, s_errno, strerror(s_errno));
   }
 }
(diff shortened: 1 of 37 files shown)


src/eventer/eventer_ports_impl.c
```diff
@@ -118,9 +116,8 @@ static void alter_fd_dissociate(eventer_t e, int mask, struct ports_spec *spec)
   if (ret == -1) {
     if(s_errno == ENOENT) return; /* Fine */
     if(s_errno == EBADFD) return; /* Fine */
-    mtevL(eventer_err,
+    mtevL(mtev_error,
```
aidancully: should be `mtevFatal`

src/utils/mtev_log.h
```diff
@@ -172,6 +172,11 @@ API_EXPORT(int)
   mtevL(ls, "[FATAL] " args); \
   abort(); \
 } while(0)
+#define mtevAssert(condition) do {\
+  if(!(condition)) { \
+    mtevFatal(mtev_error, "%s\n", #condition);\
+  }\
+} while(0)
```
aidancully: should we have two variants of this?
```
#ifdef NDEBUG
#define mtevAssert(condition) do {} while(0)
#else
#define mtevAssert(condition) do { \
    if(!(condition)) mtevFatal(mtev_error, "%s\n", #condition); \
  } while(0)
#endif
```
?

aidancully: also, the mtevFatal string should be something like `mtevAssert(mtev_error, "assertion (%s) at %s:%d failed\n", #condition, __FILE__, __LINE__);`

aidancully:
LGTM

PR #79, opened 2016-06-23
Add Error Message If NDEBUG Is Defined

pamaddox:
If NDEBUG is defined, add an #error that audit of mtevAssert is
required.


aidancully:
LGTM

PR #81, opened 2016-06-24
add fd back to spec->epoll_fd when doing implicit eventer_add during …

aidancully:
…eventer_trigger.

seems to fix case where we hit an assertion that EPOLL_CTL_MOD failed due
to the fd not being associated to the epoll_fd.


postwait:
I dont think we should act there. We should act where we mod and choose to add when appropriate

aidancully:
OK, that occurred to me, too... The logic seemed more difficult, since we'd need to either `ADD` instead of `MOD`, or do nothing instead of `DEL`... but I can take that approach if you prefer.

PR #84, opened 2016-06-30
Add 2048-bit DH params support

esproul:
Removes 512-bit DH params, which are now considered insecure.
Use the strongest params available for new connections.
Add some logging around params loading and generating.

### DIFF
--- a/src/eventer/eventer_SSL_fd_opset.c
+++ b/src/eventer/eventer_SSL_fd_opset.c
@@ -172,7 +172,7 @@ load_dh_params(const char *filename) {
   if(filename == NULL) return NULL;
   bio = BIO_new_file(filename, "r");
   if(bio == NULL) return NULL;
-  mtevL(mtev_error, "Loading DH parameters from %s.\n", filename);
+  mtevL(mtev_notice, "Loading DH parameters from %s.\n", filename);
   PEM_read_bio_DHparams(bio, &dh, 0, NULL);
   BIO_free(bio);
   if(dh) {
@@ -195,7 +195,7 @@ save_dh_params(DH *p, const char *filename) {
   if(fd < 0) return;
   bio = BIO_new_fd(fd, 0);
   if(bio == NULL) { close(fd); return; }
-  mtevL(mtev_error, "Saving DH parameters to %s.\n", filename);
+  mtevL(mtev_notice, "Saving DH parameters to %s.\n", filename);
   PEM_write_bio_DHparams(bio,p);
   BIO_free(bio);
   fchmod(fd, 0400);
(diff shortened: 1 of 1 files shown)


src/eventer/eventer_SSL_fd_opset.c
```diff
@@ -791,7 +792,9 @@ eventer_ssl_ctx_new(eventer_ssl_orientation_t type,
   }
 
   ctx->ssl = SSL_new(ctx->ssl_ctx);
-  if(dh512_tmp && dh1024_tmp)
+  if(dh2048_tmp)
+    SSL_set_tmp_dh_callback(ctx->ssl, tmp_dh_callback);
+  if(dh1024_tmp)
```
postwait: This should be one if and one action.   If (a || b) thing

src/eventer/eventer_SSL_fd_opset.c
```diff
@@ -1115,18 +1118,18 @@ void eventer_ssl_init() {
...
+  if (!dh1024_file || strcmp(dh1024_file, "")) {
     e = eventer_alloc();
     e->mask = EVENTER_ASYNCH;
     e->callback = generate_dh_params;
-    e->closure = (void *)512;
+    e->closure = (void *)1024;
     eventer_add_asynch(NULL, e);
   }
-  if (!dh1024_file || strcmp(dh1024_file, "")) {
+  if (!dh2048_file || strcmp(dh2048_file, "")) {
```
postwait: I thought of we didn't specify a 2048 file,  we were going to skip gen

postwait:
I don't see many error messages. But they are labeled as mtev_error... they should be eventer debug directed.

src/eventer/eventer_SSL_fd_opset.c
esproul: Since the DH param generation is non-blocking, and since you'll at least have 1024-bit params in < 1 second, I felt we could let the 2048-bit generation proceed.  It doesn't appear to hurt anything, and as soon as its done, new DHE connections will get the larger params.  We will be shipping a 2048-bit params file with our broker package, so we should have the larger params available immediately.

esproul:
I wanted visibility into DH param loading/generation in the default logs, and eventer debug is of course turned off by default.  I feel operators should not have to run in debug to see the timing of generation and/or whether params are being loaded from a file.

src/eventer/eventer_SSL_fd_opset.c
esproul: I will fix this.

src/eventer/eventer_SSL_fd_opset.c
postwait: So to avoid the param generation completely I must specify a file whose name is "" ?

esproul: 2048-bit works just like 1024-bit does today.  We've not had a way to completely disable generating a given set of DH params, AFAIK.

src/eventer/eventer_SSL_fd_opset.c
```diff
@@ -172,6 +172,7 @@ load_dh_params(const char *filename) {
   if(filename == NULL) return NULL;
   bio = BIO_new_file(filename, "r");
   if(bio == NULL) return NULL;
+  mtevL(mtev_notice, "Loading DH parameters from %s.\n", filename);
```
postwait: I still think this should be debugging... if it fails, error.

I can agree that non-normatives (like generating it and saving it) could be considered notices.  But loading the params isn't notable.

src/eventer/eventer_SSL_fd_opset.c
postwait: We haven't had a way to disable them because they were so nominally cheap...

And it would appear that setting it to "" will disable the generation (AFAICT).  That's acceptable.

src/eventer/eventer_SSL_fd_opset.c
esproul: So you would be OK with making the loading message `eventer_deb`?  If so, I'm happy to do that.

postwait: The goal is: if everything is working normally I should see nothing.  If I'm doing something unique and expensive, perhaps a notice, and if fuck something up I should definitely get an error -- otherwise: debug.

postwait:
LGTM

PR #85, opened 2016-07-01
mtev_http_session_req_consume shouldn't return 0 when it starts readi…

aidancully:
…ng an empty bchain buffer.

mtev_http_request_finalize can sometimes leave an empty bchain
buffer as the only buffer on the bchain. It had been the case
that mtev_http_session_req_consume would, in this case, return 0,
but better to actually try another non-blocking read, instead.


pamaddox:
Looks fine to me.

PR #86, opened 2016-07-06
release http context as soon as we detect a socket error.

aidancully:
Existing code would possibly perform websockets check after the
file-descriptor for an HTTP connection was closed. Better to
just free the connection structure immediately when an error is
detected.

### DIFF
--- a/src/mtev_http.c
+++ b/src/mtev_http.c
@@ -1435,6 +1435,7 @@ mtev_http_session_drive(eventer_t e, int origmask, void *closure,
     mask = mtev_http_complete_request(ctx, origmask);
     mtevL(http_debug, "   <- mtev_http_complete_request(%d) = %d\n",
           e->fd, mask);
+    if(ctx->conn.e == NULL) goto release;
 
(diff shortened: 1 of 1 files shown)


src/mtev_http.c
```diff
@@ -1449,7 +1450,6 @@ mtev_http_session_drive(eventer_t e, int origmask, void *closure,
     } else {
 #endif
       _http_perform_write(ctx, &maybe_write_mask);
-      if(ctx->conn.e == NULL) goto release;
```
postwait: are you sure this should be removed.  I thought _http_perform_write could potentially destroy the connection and that was why it was immediately checked after this call.

aidancully: I read through that code, and it looked like it would never remove `conn.e`.

aidancully: I can easily put that line back, I agree it's not obvious that `_http_perform_write` will _not_ release `conn.e`.

postwait:
LGTM

PR #88, opened 2016-07-07
fix SPMC in mtev_hash

rileyberton:


postwait:
Looks good to me.  Do any of the other c tests set examples for output?  Tap perhaps?

src/utils/mtev_hash.h
```diff
@@ -38,11 +38,25 @@
...
+#include <ck_spinlock.h>
 
 typedef void (*NoitHashFreeFunc)(void *);
 
-typedef struct {
+typedef enum mtev_hash_lock_mode {
+  MTEV_HASH_LOCK_MODE_NONE = 0,
+  MTEV_HASH_LOCK_MODE_MUTEX = 1,
+  MTEV_HASH_LOCK_MODE_SPIN = 2,
+  MTEV_HASH_LOCK_MODE_SIZE
```
aidancully: why add MTEV_HASH_LOCK_MODE_SIZE? on some compilers, it'll generate a warning in the `switch` in `set_lock_mode_funcs` that some enum values aren't handled.

rileyberton:
I don't believe there are any other C tests in mtev at the moment.  they are all lua.

src/utils/mtev_hash.h
rileyberton: Actually, this is not needed, it was there for an earlier approach where lock functions we stored in an array.  I will remove.

src/utils/mtev_hash.c
```diff
@@ -189,7 +228,33 @@ struct ck_hs_map {
...
+{
+  switch (lock_mode) {
+  case MTEV_HASH_LOCK_MODE_NONE:
+    h->lock = &none_lock;
+    h->unlock = &none_unlock;
+    break;
+  case MTEV_HASH_LOCK_MODE_MUTEX:
+    pthread_mutex_init(&h->locks.hs_lock, NULL);
+    h->lock = &mutex_lock;
+    h->unlock = &mutex_unlock;
```
aidancully: do we need to worry about where a corresponding `pthread_mutex_destroy` would go?

src/utils/mtev_hash.c
```diff
@@ -200,7 +273,16 @@ void mtev_hash_init_size(mtev_hash_table *h, int size) {
...
                          size, lrand48()));
   mtevAssert(h->hs.hf != NULL);
+
+  mtev_hash_set_lock_mode_funcs(h, lock_mode);
+}
+
+void
+mtev_hash_set_lock_mode(mtev_hash_table *h, mtev_hash_lock_mode_t lock_mode)
+{
+  mtev_hash_set_lock_mode_funcs(h, lock_mode);
```
aidancully: should probably call `destroy_locks` in this function, before calling `set_lock_mode_funcs`? or else `set_lock_mode_funcs` should call `destroy_locks` internally?

rileyberton: meh.. it's not really safe to alter the lock mode.. not even sure we need this function in deference to just using the proper init function

aidancully: Note that if you call the old `init` function, that defaults to `MODE_MUTEX`... so that calling this function after the old `init` will not DTRT.

aidancully: I'd also be OK with just removing this function from the public API? If you want non-mutex locks, call the new init function; and it's most appropriate to set the lock type at init time, anyway.

PR #91, opened 2016-07-13
Explicitly Initialize Mtev Hash Structures

pamaddox:
Rather than setting them with MTEV_HASH_EMPTY, explcitly call
mtev_hash_init to initialize hash tables.


aidancully:
LGTM

PR #94, opened 2016-07-15
Log When We Init Hashes In Failure Case

pamaddox:
If we try to use a hash function on an uninitialized hash table, log a
failure (including a stack trace), then initialize it. The idea is to
eventually require hashes to be initialized before use - this will aid
in the transition.


aidancully:
LGTM

PR #95, opened 2016-07-15
improve error reporting when the HTTP download exceeds the read_limit.

aidancully:


aidancully:
I should point out, I'm not at all sure that those are the best error messages to return, and would like to hear if someone has opinions about what error text would be better.

pamaddox:
This looks fine to me. I don't have any objections to the error messages being returned here.

rileyberton:
Looks good to me

PR #97, opened 2016-07-18
Reinitialize Response Hash If Freed

pamaddox:
If we've freed the response hash table, we need to reinitialize it so we
don't attempt to use a freed hash table.

Long term, this code likely needs to be refactored to avoid this
condition at all; this is a band-aid.


aidancully:
LGTM

PR #98, opened 2016-07-18
Reallocate Request Hashes If Reused

pamaddox:
Much like requests - reallocate hashes if we've already freed them.


aidancully:
LGTM

PR #99, opened 2016-07-19
Set SONAME to major.minor

esproul:
This eases patch-level, bugfix updates by getting dependent
applications to depend only on major.minor as the library filename.


esproul:
Tested that SONAME is correct:

```
$ elfdump -d /tmp/soname_test/opt/circonus/lib/amd64/libmtev.so.0.0.6 | grep SONAME
      [24]  SONAME            0xacc0              libmtev.so.0.0
```

and that an application built against it looks for the correct file:

```
    trying path=/opt/circonus/lib/amd64/libmtev.so.0.0
        libmtev.so.0.0 =>        /opt/circonus/lib/amd64/libmtev.so.0.0
```

rileyberton:
Looks good to me

pamaddox:
Looks fine to me.

PR #105, opened 2016-07-20
Implemented mtev.cluster_handle_node_update_hook

JonasKunze:
Added two functions to the lua mtev object to access cluster specific information:

cluster_handle_node_update_hook_register (called every time the cluster has changed)
cluster_am_i_oldest_node (check if the given node is currently the oldest on in the cluster)


pamaddox:
Looks good to me.

JonasKunze:
wow, that was quick! thx!

PR #106, opened 2016-07-20
Assert Threads Are Equal For mtev_lua_on_cluster_node_updated

pamaddox:
Add a pthread_t "self" to lua_callback_ref, set it to pthread_self() in
nl_cluster_handle_node_update_hook_register, then confirm that it's the
same in the callback (currently only mtev_lua_on_cluster_node_updated).


aidancully:
:+1:

PR #107, opened 2016-07-21
Support mtev programs that avoid mtev_main and still need init_globals.

postwait:
Make mtev_init_globals non-static.
Make it callable multiple times safely.
Make it called in the .init section if libmtev is linked shared.


rileyberton:
Looks good to me

PR #108, opened 2016-07-21
Implemented mtev.cluster_details

JonasKunze:
Replaced `cluster_handle_node_update_hook_register` and `cluster_am_i_oldest_node` with `cluster_details` and `cluster_get_self`

### DIFF
--- a/src/modules/lua_mtev.c
+++ b/src/modules/lua_mtev.c
@@ -124,8 +124,14 @@ typedef struct {
 static void
 mtev_lua_push_timeval(lua_State *L, struct timeval time) {
   double seconds;
-  seconds = time.tv_sec + time.tv_usec / 1000000.0;
-  lua_pushnumber(L, seconds);
+  lua_getglobal(L, "mtev");
+  lua_getfield(L, -1, "timeval");
+  lua_getfield(L, -1, "new");
+  lua_replace(L, -3); // replaces mtev with new and removes new
+  lua_pop(L, 1); // pops timeval
+  lua_pushinteger(L, time.tv_sec);
+  lua_pushinteger(L, time.tv_usec);
+  lua_call(L, 2, 1);
 }
 static void
 nl_extended_free(void *vcl) {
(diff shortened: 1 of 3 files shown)


src/modules/lua_mtev.c
```diff
@@ -129,6 +122,12 @@ typedef struct {
 } json_crutch;
 
 static void
+mtev_lua_push_timeval(lua_State *L, struct timeval time) {
```
postwait: We should be using mtev.timeval for this...

`mtev.timeval.new(sec, usec)`

src/modules/lua_mtev.c
```diff
@@ -3419,120 +3417,125 @@ nl_watchdog_child_heartbeat(lua_State *L) {
...
+  mtev_lua_setup_cluster_node(L, mtev_cluster_get_oldest_node(cluster));
+  lua_settable(L, -3);
 
-  lua_call(cb_ref->L, 3, 1);
-  rv = lua_tointeger(cb_ref->L, -1);
+  lua_pushstring(L, "nodes");
+  lua_createtable(L, number_of_nodes, 0);
+  for(i=0; i != number_of_nodes; ++i) {
+    lua_pushinteger(L, i+1);
+    mtev_lua_setup_cluster_node(L, nodes[i]);
```
postwait: This node[i] could be freed after return but before use in lua. We need a "dumb lua" copy here.

src/modules/lua_mtev.c
JonasKunze: mtev.timeval.new is lua code that generates a timeval, right?
here I want to go the other way round: push a double to lua
I'm doing this because:
1) I saw other code doing this
2) I didn't see any lua code handling a timeval directly

In principle we could also implement a `setup_timeval` that pushes a userdata timeval with the corresponding __index function, but I think a double is fine.

postwait:
LGTM

src/modules/lua_mtev.c
JonasKunze: I'm using mtev.timeval.new now

PR #109, opened 2016-07-21
Add Locking To Certain Hash Tables

pamaddox:
Did an audit to initialize certain hash tables as locking. I didn't mess
with any tables that manually have locks around them, and I tried to be
conservative with initializing the locks so that we don't lock
unnecessarily - this may require tweaking at a later point, but this is
a good start.

### DIFF
--- a/src/eventer/eventer.c
+++ b/src/eventer/eventer.c
@@ -191,8 +191,8 @@ int eventer_choose(const char *name) {
 }
 
 void eventer_init_globals() {
-  mtev_hash_init_locks(&__name_to_func, 256, MTEV_HASH_LOCK_MODE_MUTEX);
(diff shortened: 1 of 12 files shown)


aidancully:
mtev_hash.c uses `NoitHASH_INITIAL_SIZE` as the initial size, `1<<7`, or 128... This patch changed those all to 256, which will increase our memory usage. Not a big problem, but... Can we add a `#define MTEV_HASH_DEFAULT_SIZE (1<<7)` to `mtev_hash.h`, and change those 256's to use that constant, and replace all `NoitHASH_INITIAL_SIZE` with `MTEV_HASH_DEFAULT_SIZE`?

aidancully:
LGTM

PR #110, opened 2016-07-21
mtev_config.h will export uint<8|16|32>_t now, following stdint.h convention.

aidancully:


postwait:
I'm fine with this in concept... libmtev supports a lot of platforms... so we need to test this on linux old-new, solaris 10+, freebsd 8-11 and mac os x at a minimum.

aidancully:
Tested on rhel6.3 (old linux), yesterday's Arch Linux (new linux), omnios-006, and omnios-014. (Verified that mtev_config.h looked as expected, and that it was possible to compile something `#include`ing `<mtev_config.h>` with a uint8_t defined.) I'll test on FreeBSD tomorrow.

aidancully:
I don't believe libmtev currently compiles on FreeBSD, due (at least) to `mtev_hrtime` not currently being ported to the platform. I can't test on Mac OS X easily, but I strongly doubt, at this point, that I've introduced any portability issues with this PR.

postwait:
+1

PR #111, opened 2016-07-26
Werror

rileyberton:
-Werror was not on for libmtev and there were many warnings (mostly copy/paste unused variable).

Pay particular attention to `mtev_net_heartbeat.c` which defined and filled an IPV6 struct but then never used it?


aidancully:
Can you run `mtev_config --cflags` with this change? I'm worried `-Werror` will be included in the output, and I think it shouldn't be.

rileyberton:
Yes, mtev-config does the wrong thing now.  looking into it.

postwait:
needs work... doing that now.

src/modules/mtev_fq.c
```diff
@@ -323,7 +323,7 @@ fq_driver_config(mtev_dso_generic_t *img, mtev_hash_table *options) {
   void *data;
 
   while(mtev_hash_next(options, &iter, &key, &klen, &data)) {
-    mtevL(mtev_error, "%s %s!\n", key, data);
+    mtevL(mtev_error, "%s %s!\n", key, (char *)data);
```
aidancully: would it be better to use `mtev_hash_next_str` here, instead of adding a typecast?

postwait:
LGTM

PR #117, opened 2016-07-29
Implemented mtev.conf_get_string_list

JonasKunze:
Returns a table with the xpath matching <arg2> of all nodes matching <arg1>

Example:
<a>
  <b c="attr">value</b>
  <b c="attr2" />
</a>

mtev.conf_get_string_list("/a/b", "@c")
  -> {1 = "attr", 2 = "attr2"}

mtev.conf_get_string_list("/a/b", "self::node()")
  -> {1 = "value", 2 = ""}


rileyberton:
Looks good to me

PR #121, opened 2016-08-02
Bugfix: Shortly dead cluster nodes became oldest node

JonasKunze:
If a cluster node that was the oldest node died and came up within a short time (shorter then the timeout) it became oldest node again.


postwait:
LGTM

PR #122, opened 2016-08-02
Enabled appending of arbitrary payload to cluster heartbeats

JonasKunze:
Also the config_seq is stored in the node struct and passed to the user (he should decide what to do if the configs diverged)

### DIFF
--- a/src/mtev_cluster.c
+++ b/src/mtev_cluster.c
@@ -51,6 +51,25 @@ static const struct timeval boot_time_of_dead_node = { (1UL
     << (CHAR_BIT * sizeof(boot_time_of_dead_node.tv_sec) - 1)) - 1, (1UL
     << (CHAR_BIT * sizeof(boot_time_of_dead_node.tv_usec) - 1)) - 1 };
 
+#define HEART_BEAT_HDR_LEN 1 + UUID_SIZE + sizeof(u_int64_t) + sizeof(u_int64_t) + 1
+#define MAX_PAYLOAD_LEN_SUM  1518 - 14/*ETH*/ - 20 /*IP*/ - HEART_BEAT_HDR_LEN - 4 /*FCS*/
+
+#define HEARTBEAT_MESSAGE_VERSION 1
+static const uint8_t HEARTBEAT_MESSAGE_VERSION_AND_UNDERSTOOD = (HEARTBEAT_MESSAGE_VERSION << 4) | HEARTBEAT_MESSAGE_VERSION;
+
+typedef struct {
+  uint8_t app_id;
+  uint8_t key;
+  uint8_t data_len;
+} __attribute__ ((__packed__)) hb_payload_hdr_t;
+
+typedef struct {
+  uint8_t app_id;
+  uint8_t key;
+  uint8_t data_len;
+  void* data;
+} hb_payload_t;
+
 /* All allocated with mtev_memory_safe commands */
 struct mtev_cluster_t {
   char *name;
@@ -65,13 +84,17 @@ struct mtev_cluster_t {
   mtev_cht_t **cht;
   mtev_net_heartbeat_ctx *hbctx;
   mtev_cluster_node_t *oldest_node;
+  mtev_hash_table hb_payloads;
 };
 
+
 MTEV_HOOK_IMPL(mtev_cluster_handle_node_update,
-  (mtev_cluster_node_t *updated_node, mtev_cluster_t *cluster),
+  (mtev_cluster_node_t *updated_node, mtev_cluster_t *cluster,
+      struct timeval old_boot_time),
   void *, closure,
-  (void *closure, mtev_cluster_node_t *updated_node, mtev_cluster_t *cluster),
-  (closure,updated_node,cluster))
+  (void *closure, mtev_cluster_node_t *updated_node, mtev_cluster_t *cluster,
+      struct timeval old_boot_time),
+  (closure,updated_node,cluster,old_boot_time))
 
 static int
 mtev_cluster_node_compare(const void *a, const void *b) {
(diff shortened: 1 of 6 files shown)


postwait:
I believe this should be marginally more flexible.

As single payload is challenging.  I suggest payloads be marked with a 32bit int "Type" and a unsigned 8bit "size"...

The payload you described with cluster config can be of one type and would be encoded into the packet as <32bit type><8bit len><len....>.

This will allow libmtev to share some stuff (like cluster state) and reconnoiter could add on it's own small amount of metadata...

the 8bit size forces callers to recognize this is going into a small packet and it isn't supposed to be used as a transport as much as a notification of state change.

src/mtev_cluster.c
```diff
@@ -65,13 +82,16 @@ struct mtev_cluster_t {
   mtev_cht_t **cht;
   mtev_net_heartbeat_ctx *hbctx;
   mtev_cluster_node_t *oldest_node;
+  struct jl_array_list *hb_payloads;
```
postwait: I think ti is very awkward to use an internal json list implementation here.  Can't we just use a normal C array? or an mtev_hash ?

JonasKunze: The only JS specific line I see in mtev_arraylist is `mtev_json_max`. So I don't see how these array lists are not generic enough to be used for this purpose.
Using a C array would mean to implement exactly what's currently in `jl_array_list`
I didn't want to use an mtev_hash because I thought it could become a performance issue.

postwait: If we toss json-lib for an alternative implementation, those structs would go away.. I'd rather not see reliance on those structures.  I see no reason for the arrays to every shrink... so this seems like a trivial use of plain C arrays...  find first null to insert, grow to accommodate, null to remove.  Also, I do not think the  mtev_hash will be poor performance.

src/mtev_cluster.c
```diff
@@ -236,22 +268,52 @@ mtev_cluster_info_compose(void *payload, int len, void *c) {
...
   /* TODO: Support registered composers */
-  MEMWRITE_RETURN;
+  return MEMWRITE_WRITTEN;
+}
+static void
+mtev_cluster_store_payload(mtev_cluster_node_t *node, const void* payload, uint8_t payload_length) {
+  if(node->payload) {
+    free (node->payload);
+  }
+  node->payload = malloc(payload_length);
```
postwait: This seems an unnecessary free and alloc.  If you only do this on a size increase, you'll quickly converge on a situation never requiring more memory ops.

src/mtev_cluster.c
```diff
@@ -680,6 +745,114 @@ mtev_cluster_do_i_own(mtev_cluster_t *c, void *key, size_t klen, int w) {
...
+  }
+
+  if(payload_len_sum + payload_length > MAX_PAYLOAD_LEN_SUM) {
+    return mtev_false;
+  }
+
+  if(old_idx != -1) {
+    hb_payload = jl_array_list_get_idx(cluster->hb_payloads, old_idx);
+  } else {
+    hb_payload = malloc(sizeof(hb_payload_t) + sizeof(payload));
```
postwait: where is this freed?

I think the caller should responsible for providing the memory and a free function.

JonasKunze: It's freed at line 810

please note that this is hb_payload does not store the payload itself but just a pointer to it! every time a heartbeat message is generated mtev_cluster copies the data from there to the actual message. this way the user doesn't have to inform mtev_cluster that the payload has changed.

src/mtev_cluster.c
JonasKunze: You're absolutely right! Should be fixed now. Will be tested tomorrow with other changes to come

src/mtev_cluster.c
JonasKunze: What do you mean by `I see no reason for the arrays to every shrink`? I don't see where `jl_array_list` shrinks. 
So if this `mtev_arraylist` anything different to what you typically expect from an arraylist, I'd suggest to rename this to e.g. `mtev_json_list` and maybe create a new `mtev_arraylist` which does exactly what you suggested. But before I do that I'd really like to understand what's so special about the current implementation. The only thing I'd change is not storing an array of pointers but an array of structs to have a better cash locality.

postwait: json-lib is third-party code that we've name-spaced... but we try to keep the changes to an absolutely minimum so that we can keep it up to date with the 3rdparty upstream.  Renaming this isn't an option.  Reimplementing this to be part of utils is an option, but I personally am not a fan of convenience functions for the most simple of data structure manipulations.

src/mtev_cluster.c
```diff
@@ -223,35 +253,80 @@ mtev_cluster_check_timeouts() {
...
+    while(mtev_hash_next(&cluster->hb_payloads, &payload_iter, &key, &klen, &vdata)) {
+      hb_payload = vdata;
+      MEMWRITE(&hb_payload->app_id, 1);
+      MEMWRITE(&hb_payload->key, 1);
+      MEMWRITE(&hb_payload->data_len, 1);
+    }
+    // write payload
+    while(mtev_hash_next(&cluster->hb_payloads, &payload_iter2, &key, &klen, &vdata)) {
+      hb_payload = vdata;
+      MEMWRITE(hb_payload->data, hb_payload->data_len);
```
postwait: This is confusing... you write out all the headers first, then all the payloads?  This is atypical serialization.  Shouldn't it be one iter loop that writes (1,1,1,data_len) bytes, followed by another of the same?  How would you know where all the sequential headers stop and the data starts?

src/mtev_cluster.c
```diff
@@ -40,6 +40,7 @@
 #include "mtev_cluster.h"
 #include "mtev_cht.h"
 #include "mtev_net_heartbeat.h"
+#include "mtev_arraylist.h"
```
postwait: what is this? and what is it used for?

JonasKunze: before I've switched to using mtev_hash I've used these arraylists to store the payloads, remember? I just forgot to remove this line, sorry!

src/mtev_cluster.c
JonasKunze: That's actually very typical for my experience. It's a similar concept to the common rule for high performance: Use structs of arrays instead of arrays of structs to reduce cash misses.

The concept here is that you have a very compact list of headers fitting into one page. You can quickly iterate over it to search for the header you're looking for while summing up the payload lengths. Once your sum is larger than the total size of your data block you know that you've reached the end of the header list. Otherwise the sum points to the payload. This is done here:  https://github.com/JonasKunze/libmtev/blob/smart_cluster/src/mtev_cluster.c#L815

You can argue that a protocol with a high cash locality is not important here as we're anyway not focusing on performance, but that's simply how I've learned to design binary protocols many times.

If you want, I can change this implementation to what you've suggested.

src/mtev_cluster.c
```diff
@@ -223,54 +255,118 @@ mtev_cluster_check_timeouts() {
...
   boot_time.tv_usec = (packed_time & 0xffffff);
   MEMREAD(&seq, sizeof(seq));
   seq = ntohll(seq);
+  MEMREAD(&number_of_payloads, 1);
+
+  payload_len = len-MEMREAD_BYTES_READ;
+  MEMGET(payload, payload_len);
+
+  // We currently support only one version
+  assert (version == understood_version && version == HEARTBEAT_MESSAGE_VERSION);
```
postwait: Another Node could crash us. If they aren't the same we should just ignore the packet (and log)

JonasKunze: agreed

PR #130, opened 2016-08-18
Passing change type to handle_node_update_hook

JonasKunze:
Added an enum defining the possible reason an `node_update_hook` has been called. The corresponding value is passed to the hook.

### DIFF
--- a/src/mtev_cluster.h
+++ b/src/mtev_cluster.h
@@ -56,12 +56,11 @@ typedef struct {
   uint8_t number_of_payloads;
 } mtev_cluster_node_t;
 
-typedef enum {
-  mtev_cluster_node_died,
-  mtev_cluster_node_rebooted,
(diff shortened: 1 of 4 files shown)


src/mtev_cluster.h
```diff
@@ -56,6 +56,13 @@ typedef struct {
   uint8_t number_of_payloads;
 } mtev_cluster_node_t;
 
+typedef enum {
+  mtev_cluster_node_died,
+  mtev_cluster_node_rebooted,
+  mtev_cluster_node_changed_seq,
+  mtev_cluster_node_changed_payload
+} mtev_cluster_node_changes;
```
postwait: We typically all caps enum elements.

Also, it seems that more than one of these could be true at the same time. Would a bit set be more appropriate?

postwait:
LGTM

PR #131, opened 2016-08-18
Thread binding changes

rileyberton:
* do not bind the mtev gc thread, asynch log thread or the eventer threads
* leave jobq threads bound to cores
* provide an env var MTEV_THREAD_BINDING_DISABLE which allows turning off binding for everything


aidancully:
LGTM

PR #133, opened 2016-08-29
Made timeval structs const in mtev_log functions

JonasKunze:


src/utils/mtev_log.c
```diff
@@ -168,7 +168,7 @@ intersect_seg(int a1, int a2, int b1, int b2) {
   return rv;
 }
 static int
-membuf_logio_writev(mtev_log_stream_t ls, const struct timeval *whence,
+membuf_logio_writev(mtev_log_stream_t ls, const const struct timeval *whence,
```
postwait: twice the const isn't twice as good.

postwait:
Can we identify whoe implements these hooks today? All their code will break as their hook implementations will have prototype mismatches.

JonasKunze:
I only found one place:
https://github.com/circonus-labs/reconnoiter/blob/5b936e6c0a76730aa4359d8ca9c5f9b882827a40/src/noit_metric_director.c#L291

If you think it's a bad Idea to change this hook I can change my code base instead to use non-const timevals...

postwait:
LGTM

PR #138, opened 2016-09-02
Implemented mtev_str_buff

JonasKunze:
Just added an exponentially growing (base 2) buffer for string concatenation similar to the already existing prepend_str


rileyberton:
Looks good to me

PR #139, opened 2016-09-02
Refactor fast time support.

postwait:
* Use a single global rdtsc function when found.
 * Store skew and tpn together in the coreclock.
 * Calculate hrtime to gettimeofday skew: mtev_gettimeofday.
 * Tracks the number of fast and slow calls.
 * Allow runtime toggling.
 * Automatic per-cpu killswitch if lack of maintenance is
   determined.
 * Add dedicated maintenance thread for tpn and skew upkeep.
 * Add console "mtev rdtsc <enable|disable>"
 * Add console "show mtev debug coreclocks"

Now all threads benefit if they find themselves on a cpu
that has been evaluated "accurately enough" and "recently
enough."  On illumos this is a 300ns -> 90ns speed up for
the common case.


rileyberton:
Looks good to me

PR #142, opened 2016-09-09
Fix Memory Leak When HTTP Connection Snips

pamaddox:
Need to clean up the service if the connection gets broken.


postwait:
LGTM

PR #147, opened 2016-10-17
Fix memory leak in abandoned gunzip.

postwait:
Change 175e7c09 introduced a refactor that didn't propagate the
unconditional free into all switch cases.


aidancully:
LGTM

PR #152, opened 2016-11-08
Respect CN as a unique key for connection initiation filtering.

postwait:
It is possible to have multiple noits without an address.  If
that was the case and we initiated the connection, we'd start
another (extra) connection for all address-less destinations.
We should expect CNs to be unique, so allow the caller to specify
that for filtering as well.  No CN and IP address should be
coincidentally the same, so this is safe without breaking existing
consumers of the API.


rileyberton:
Looks good to me

PR #153, opened 2016-11-08
Fix the lua1.conf and luatest.lua to run.

postwait:
This embeds all tests as a part of the example program.
Multi-threaded tests don't work yet due to lack of
cross-thread notify/waitfor support.


aidancully:
This seems to have introduced some code duplication? For example, I see both src/examples/lua-tests/confstr_spec.lua, which looks very similar (if not identical) to test/utils/confstr_spec.lua.

postwait:
The src/examples/lua-test/ tests are a superset of the the test/utils/ tests.  Once this is all working we can remove them from tests.... or move this all to tests.. or something.  Those that are in both are identical.

aidancully:
LGTM

PR #154, opened 2016-11-08
first version of websocket client

gallison93:


src/mtev_websocket_client.c
```diff
@@ -0,0 +1,461 @@
...
+#ifdef HAVE_WSLAY
+  int fd = -1, rv;
+  struct addrinfo *res = NULL, *res_ptr = NULL, hints;
+  char port_str[6];
+
+  snprintf(port_str, 6, "%d", port);
+  memset(&hints, 0, sizeof(hints));
+  hints.ai_family = AF_UNSPEC;
+  hints.ai_socktype = SOCK_STREAM;
+  if((rv = getaddrinfo(host, port_str, &hints, &res))) {
```
postwait: This is a blocking call. I would have this function just take an IPv4 or IPv6 address -- no lookup.  The caller can struggle with the non-blocking lookup.

src/mtev_websocket_client.c
```diff
@@ -0,0 +1,461 @@
...
+      client->sent_handshake = mtev_true;
+    }
+    
+    if(client->sent_handshake && mask & EVENTER_READ) {
+      if(mtev_websocket_client_recv_handshake(client) == mtev_false) {
+        mtevL(mtev_error, "mtev_websocket_client_recv_handshake failed, aborting drive\n");
+        goto abort_drive;
+      }
+      wslay_event_context_client_init(&client->wslay_ctx, &wslay_callbacks, client);
+      client->did_handshake = mtev_true;
```
postwait: I would add a client call back here (and in the case) so the programmer can react to the completion "ready state" of the websockets connection.

src/mtev_websocket_client.c
```diff
@@ -0,0 +1,461 @@
...
+    eventer_remove_fd(client->e->fd);
+    client->e->opset->close(client->e->fd, &mask, client->e);
+    eventer_free(client->e);
+    if(client->did_handshake)
+      wslay_event_context_free(client->wslay_ctx);
+    free((void *)client->path);
+    free((void *)client->service);
+    free((void *)client->host);
+    client->closed = mtev_true;
+  }
```
postwait: I'd add a callback here.  So the programmer can react.

rileyberton:
Looks good to me

PR #160, opened 2016-11-16
Introduce mtev_hash_adv (obsolete mtev_hash_next{,_str})

postwait:
It has been, but will no longer be, tragic that each attempt
to use an iterator over an mtev_hash required the stack declarations
of at least a const char *key, an int klen, and a void *value.
This commit adds those requirements into the iterator type itself
and preserves backward compatibility.  All old calls will only access
within the boundaries of its original size so while the the ABI of the
iterator changed, it remains safe and backward compatible.


rileyberton:
Looks good to me

PR #161, opened 2016-11-18
Implicitly create logs when searching for them.

postwait:
Now if you ask for a specific log, it will always come back.  This
also plays some type trickery with #define so that mdb can easily
print mtev_log_stream_t objects: `::mtev_log | ::pring mtev_log_stream_t`

### DIFF
--- a/src/utils/mtev_log.c
+++ b/src/utils/mtev_log.c
@@ -1325,7 +1325,7 @@ mtev_log_final_resolve() {
   return mtev_true;
 }
 
-mtev_log_stream_t
+static mtev_log_stream_t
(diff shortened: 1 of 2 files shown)


src/utils/mtev_log.c
```diff
@@ -1326,9 +1326,9 @@ mtev_log_final_resolve() {
 }
 
 mtev_log_stream_t
-mtev_log_stream_new(const char *name, const char *type, const char *path,
-                    void *ctx, mtev_hash_table *config) {
-  mtev_log_stream_t ls, saved;
+mtev_log_stream_new_internal(const char *name, const char *type, const char *path,
```
aidancully: should this function be `static`?

PR #162, opened 2016-11-18
initial implementation of alternative event loop pools

postwait:


src/eventer/eventer_impl.c
```diff
@@ -103,16 +104,49 @@ eventer_impl_t __eventer = NULL;
...
+}
+
+eventer_pool_t *eventer_pool(const char *name) {
+  void *vptr;
+  if(mtev_hash_retrieve(&eventer_pools, name, strlen(name), &vptr))
+    return (eventer_pool_t *)vptr;
+  return NULL;
+}
+
+int eventer_loop_concurrency() { return __total_loop_count; }
```
aidancully: I think this should return `default_pool.__loop_concurrency`, and we should have a new routine, `eventer_loop_concurrency_pool` to return the concurrency for a specific pool?

src/eventer/eventer_impl.c
```diff
@@ -306,13 +361,17 @@ void eventer_loop() {
...
+        pool->name, pool->__loop_concurrency);
+  for(i=start; i<pool->__loop_concurrency; i++) {
     pthread_t tid;
-    pthread_create(&tid, NULL, thrloopwrap, (void *)(vpsized_int)i);
+    int adjidx = pool->__global_tid_offset + i;
+    mtevAssert(pool == eventer_impl_tls_data[adjidx].pool);
+    pthread_create(&tid, NULL, thrloopwrap, (void *)(vpsized_int)adjidx);
   }
-  while(__loops_started < __loop_concurrency);
+  while(ck_pr_load_32((uint32_t *)&pool->__loops_started) < ck_pr_load_32((uint32_t *)&pool->__loop_concurrency));
```
aidancully: type-casts seem unnecessary?

PR #164, opened 2016-11-21
Support multiple concurrency watchdog timers.

postwait:
Since making the event loop multi-threaded, it is possible for
a thread other than the master thread to hang without triggering
a watchdog timeout.  This provides an API where any consumer
can create a new watchdog and maintain it such that the parent
will notice inactivity.  This commit does not enable per-thread
watchdogs in the eventer as that change would conflict with some
other in-flight changes; it will come later.

### DIFF
--- a/src/utils/mtev_watchdog.c
+++ b/src/utils/mtev_watchdog.c
@@ -59,10 +59,27 @@
 #include "mtev_time.h"
 #include "mtev_watchdog.h"
 
+struct mtev_watchdog_t {
+  int ticker;
+  enum {
+    CRASHY_NOTATALL = 0,
+    CRASHY_CRASH = 0x00dead00,
+    CRASHY_RESTART = 0x99dead99
+  } action;
+  enum {
+    HEART_ACTIVE_OFF = 0,
+    HEART_ACTIVE_ON = 1,
+    HEART_ACTIVE_SKIP = 2
+  } active;
+  struct {
+    struct timeval last_changed;
+    int last_ticker;
+  } parent_view;
+};
+
 #define CHILD_WATCHDOG_TIMEOUT 5 /*seconds*/
-#define CRASHY_CRASH 0x00dead00
-#define CRASHY_RESTART 0x99dead99
 #define MAX_CRASH_FDS 1024
+#define MAX_HEARTS 1024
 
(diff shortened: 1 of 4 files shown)


src/utils/mtev_watchdog.c
```diff
@@ -114,45 +131,56 @@ void mtev_watchdog_ratelimit(int retry_val, int span_val) {
...
+  if(lifeline->parent_view.last_changed.tv_sec == 0) return 0;
 
-  sub_timeval(now, lastchange, &diff);
+  sub_timeval(*now, lifeline->parent_view.last_changed, &diff);
   return (unsigned long)diff.tv_sec;
 }
-static void it_ticks_crash() {
-  if(lifeline) lifeline[1] = CRASHY_CRASH;
+static void it_ticks_crash(mtev_watchdog_t *lifeline) {
+  if(lifeline == NULL) lifeline = mmap_lifelines;
```
aidancully: possibly this will be explained by a forthcoming PR, but since I'm only looking at this PR, why not remove all the
```
  if(lifeline == NULL) lifeline = mmap_lifelines;
```
statements, and just call these `it_ticks_*` functions with `mmap_lifelines`?

src/utils/mtev_watchdog.c
```diff
@@ -611,21 +661,49 @@ int update_retries(int* offset, time_t times[]) {
...
 }
 
-static int watchdog_tick(eventer_t e, int mask, void *unused, struct timeval *now) {
-  it_ticks();
+mtev_watchdog_t *mtev_watchdog_create() {
+  int i;
+  for(i=0; i<MAX_HEARTS; i++) {
+    mtev_watchdog_t *lifeline = &mmap_lifelines[i];
+    if(lifeline->active == HEART_ACTIVE_OFF) {
+      lifeline->active = HEART_ACTIVE_ON;
```
aidancully: I feel like we should default to `HEART_ACTIVE_SKIP`, and require the user to explicitly `mtev_watchdog_enable`...

postwait: The current implementation is on when used.  It make sense to me that if you went through the effort of creating a watchdog, that you wanted to use it.

PR #172, opened 2016-12-06
Allow jobqs to be configured via eventer config.

postwait:
Expose `jobq_<name>` as an eventer config property.
Values look like `concurrency[,min[,max[,safety]]]`.

Default newly (programmatically) created jobqs that
are not backqs to a concurrency of 1 so that they can
be used without a subsequeny set_concurrency call.


src/eventer/eventer_impl.c
```diff
@@ -241,6 +241,52 @@ int eventer_impl_propset(const char *key, const char *value) {
     eventer_pool_create(name, requested);
     return 0;
   }
+  if(!strncasecmp(key, "jobq_", strlen("jobq_"))) {
+    char *nv = alloca(strlen(value)+1), *tok;
+    memcpy(nv, value, strlen(value)+1);
+    const char *name = key + strlen("jobq_");
+
```
aidancully: should we
```c
if(!*name) return -1;
```? To prevent a key named `jobq_`, with no name specified.

aidancully:
LGTM

postwait:
This was merged, but for some reason the PR didn't close.

PR #175, opened 2016-12-11
Autoconf build updates.

jpeach:
A couple of updates to make the build more standard. First, modernize the `autoconf` macros we use so that you can use `autoreconf` to regenerate the build. Second, standardize on standard C fixed-width integer types and use the built-in `autoconf` macros to figure them out.


.gitignore
```diff
@@ -28,7 +28,8 @@ src/mtev_version.h
 src/man/*.8
 src/man/*.h
 src/Makefile.objs
-src/mtev_config.h
+src/config.h
+src/config.h.in
 src/scripts/mtev-config
```
postwait: The new separation of mtev_config.h into that and config.h will cause problems.  We ship mtev_config.h and it is namespaced sufficiently, but now depends upon config.h

It looks like the intended separation is to keep autoconfy things separate.  That was the original intention behind mtev_config.h and mtev_defines.h.

mtev_config.h.in should stay and mtev_defines.h should adopt the things in that were put in there as a part of this proposal.

PR #177, opened 2016-12-13
Leverage watchdogs vectors to protect all eventloop threads.

postwait:


aidancully:
Verified e->thr_owner would be set correctly (it's set to the current thread in `eventer_alloc`, which we're calling from the `per_thread_init` function), so that the watchdog poke would be scheduled on the current thread's event-loop. LGTM.

pamaddox:
Reviewed. Looks good to me.

PR #178, opened 2016-12-14
add test for hyperloglog and improve performance a bit

rileyberton:


src/utils/mtev_hyperloglog.c
```diff
@@ -62,33 +63,37 @@ mtev_hyperloglog_add(mtev_hyperloglog_t *hll, const void *data, size_t len)
...
-  c = (1 << hll->bitcount);
   m = mtev_hyperloglog_alpha(hll->bitcount) * c * c;
-  sum = 0;
 
-  for (int i=0; i < c; i++) {
-    sum += 1.0 / (1 << hll->regs[i]);
+  for (int i = 0; i < c; i++) {
+    uint8_t x = hll->regs[i];
+    *zero_register_count += x == 0;
+    int_sum += c / (1 << x);
```
aidancully: `c / (1 << x)` seems wrong if `bitcount` is low (like, imagine `bitcount == 4`, then `c == 16`, and this division would generally bias the sum to be too small). But the choice of numerator seems to factor out when calculating the `double sum`, so you could just use any constant? I haven't thought this through completely, but `1 << (sizeof(int) * 8 - hll->bitcount)` should work, and would be most accurate possible without losing precision, I think. Or failing that, just using `65536` as a numerator should be pretty good.

PR #179, opened 2016-12-14
Add a CHOTOPT for optimizatio of "hot" C files.

postwait:
Defaults to -O5, but can override with gmake CHOTOPT=-O2


rileyberton:
Looks good to me

PR #180, opened 2016-12-29
Allow pruning jobq threads to min_concurrency.

postwait:
Support less aggressive growing and periodic pruning of idle jobq threads.


rileyberton:
Looks good to me

PR #184, opened 2017-01-19
Pool watchdogs

postwait:


src/eventer/eventer_impl.c
```diff
@@ -236,12 +258,20 @@ int eventer_impl_propset(const char *key, const char *value) {
...
+    char *nv = alloca(strlen(value)+1), *tok;
+    memcpy(nv, value, strlen(value)+1);
     const char *name = key + strlen("loop_");
     if(strlen(name) == 0) return -1;
 
-    int requested = atoi(value);
+    ADVTOK; /* concurrency */
+    int requested = tok ? atoi(tok) : 0;
+    ADVTOK;
+    double hb_timeout = tok ? atof(tok) : 0;
```
aidancully: should we have error checking here? `strtod`, and verify that the string ends with `null`, and that `hb_timeout >= 0`?

postwait: we don't anywhere else (e.g. atoi).  Negative numbers would also likely act as expected... timeout immediately.  I see no unexpected behavior from the straight up atof, so I'm inclined to leave it as is.

aidancully: OK

aidancully:
LGTM

PR #185, opened 2017-01-20
Avoid Use Of alloca In http_construct_leader

pamaddox:
Rather than use alloca - have a static array of 16 char pointers instead
- use that if the amount we need is less than 16, otherwise malloc and
free.

### DIFF
--- a/src/mtev_http.c
+++ b/src/mtev_http.c
@@ -1753,6 +1753,7 @@ _http_construct_leader(mtev_http_session_ctx *ctx) {
   const char *protocol_str;
   int i;
   const char **keys;
+  char *static_key_array[16];
   mtev_boolean cl_present = mtev_false;
 
   mtevAssert(!ctx->res.leader);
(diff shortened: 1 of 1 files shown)


src/mtev_http.c
```diff
@@ -1804,6 +1811,7 @@ _http_construct_leader(mtev_http_session_ctx *ctx) {
         }
       }
     }
+    if (!cl_present && keys != (const char **)static_key_array) free(keys);
```
postwait: I don't think the !cl_present is necessary.

pamaddox: I think it might be needed. This is all enclosed in the following while loop:

while(!cl_present)

There's no guarantee that we get to the end of the loop with cl_present set. If we get to the end when "!cl_present" is true, we need to free the malloced data. Unless you're seeing something I'm not, I think it's safer to just include this check.

postwait: If keys is set and it isn't the static version, it needs to be free.  If *somehow* we get there with cl_present, I don't see how that simpler condition doesn't hold true.

pamaddox: We don't want to free it yet - we use it later. From the code:

    if (!cl_present && keys != (const char **)static_key_array) free(keys);
  }
  qsort(keys, i, sizeof(*keys), casesort);
  kcnt = i;
  for(i=0;i<kcnt;i++) {
    int vlen;
    const char *key = keys[i], *value;

We use that keys value after the loop. We don't want to free it there if we're going to break out of the loop because we want to use it.

pamaddox: Confirmed - if we remove that check, we crash.

I see you approved it; I'll go ahead and merge.

PR #186, opened 2017-01-23
Enzo c changes

gallison93:
Tweaks to websocket_client:

- Adjusting the return values of the callbacks
- Add some missing locking
- Add a closure
- Allow specifying which event pool the client should be on

### DIFF
--- a/src/examples/websocket_client.c
+++ b/src/examples/websocket_client.c
@@ -45,12 +45,13 @@ parse_cli_args(int argc, char * const *argv) {
   }
 }
 
-void websocket_ready_handler(mtev_websocket_client_t *client) {
+void websocket_ready_handler(mtev_websocket_client_t *client, void *closure) {
   mtev_websocket_client_send(client, 0x1, "Hello world!", 13);
 }
 
 int websocket_msg_handler(mtev_websocket_client_t *client, int opcode,
-                          const unsigned char *msg, size_t msg_len) {
+                          const unsigned char *msg, size_t msg_len,
+                          void *closure) {
   char buf[256];
   size_t len;
   snprintf(buf, msg_len, "%s", msg);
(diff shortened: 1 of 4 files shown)


src/mtev_websocket_client.c
```diff
@@ -316,7 +319,7 @@ mtev_websocket_client_drive(eventer_t e, int mask, void *closure, struct timeval
 
 mtev_websocket_client_t *
 mtev_websocket_client_new(const char *host, int port, const char *path, const char *service,
-                          mtev_websocket_client_callbacks *callbacks) {
+                          mtev_websocket_client_callbacks *callbacks, void *closure, const char *pool_name) {
```
postwait: It's a nit, but this has it assigned to the default pool if someone typos the name.  That seems like it might be better to control at the callsight.  What about changing this to take an `eventer_pool_t *` and requiring the caller to do the lookup (and thus handle a failed lookup)?

src/examples/websocket_client.c
```diff
@@ -45,22 +45,23 @@ parse_cli_args(int argc, char * const *argv) {
...
+mtev_boolean websocket_ready_handler(mtev_websocket_client_t *client, void *closure) {
+  return mtev_websocket_client_send(client, 0x1, "Hello world!", 13);
 }
 
-int websocket_msg_handler(mtev_websocket_client_t *client, int opcode,
-                          const unsigned char *msg, size_t msg_len) {
+mtev_boolean websocket_msg_handler(mtev_websocket_client_t *client, int opcode,
+                                   const unsigned char *msg, size_t msg_len,
+                                   void *closure) {
   char buf[256];
```
rileyberton: This is example code but it will stack overflow if someone sends a msg_len > 256

PR #188, opened 2017-01-31
Fix include writes.

postwait:
Comparing the private data instead of the container within caused
include files to have only their first child element written out.

### DIFF
--- a/src/mtev_conf.c
+++ b/src/mtev_conf.c
@@ -467,7 +467,12 @@ mtev_conf_magic_separate_includes(include_node_t **root_include_nodes, int *cnt)
         prev = NULL;
         for(n=include_nodes[i].insertion_point->children;
             n; n = n->next) {
-          if(prev && prev->_private != n->_private) {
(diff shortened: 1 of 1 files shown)


pamaddox:
LGTM

src/mtev_conf.c
```diff
@@ -467,7 +467,12 @@ mtev_conf_magic_separate_includes(include_node_t **root_include_nodes, int *cnt)
         prev = NULL;
         for(n=include_nodes[i].insertion_point->children;
             n; n = n->next) {
-          if(prev && prev->_private != n->_private) {
+          include_node_t *prev_cont = NULL, *n_cont = NULL;
+          if(prev && prev->_private)
+            prev_cont = ((mtev_xml_userdata_t *)prev->_private)->container;
+          if(n && n->_private)
```
aidancully: `n` is guaranteed to exist, you could just `if(n->_private)`

aidancully:
LGTM

PR #190, opened 2017-01-31
Require env

postwait:


docs-md/config/listeners.md
```diff
@@ -41,6 +41,8 @@ note that the listener's `type` attribute was actually set in a parent node.  Mo
 systems in libmtev will recusively merge from ancestors down to the a specimen node
 and use that result.  Here `type` is simply an attribute, so merging is just replacing.
 This node also has an `sslconfig`, but it doesn't use it, so we'll ignore that for now.
+The `require_env` attribute requires the `MTEV_CONTROL` environment variable to be set
+for this listener to be active; if unspecified, it is active.
```
pamaddox: Do you mean "if unspecified, it is inactive"? This says that it requires the MTEV_CONTROL environment variable to be set, but it will be active anyway if unspecified; regardless, documentation is unclear here.

postwait: If the require_env is unspecified, the the listener will be active.

postwait: I've reworked the docs.  And added pcre support.

rileyberton:
Looks good to me

PR #195, opened 2017-02-07
Mt lua web

postwait:
Make lua_web thread safe and allow assigning handlers to alternate event loops


rileyberton:
I don't see any obvious bugs, LGTM

PR #197, opened 2017-02-07
fix chunked encoding, make compression work with chunked encoding

rileyberton:
Please to review this carefully.

I tested using curl with every combination of chunked/not-chunked compression/not-compression.

However, this is really hairy and could use a careful review.

### DIFF
--- a/src/mtev_http.c
+++ b/src/mtev_http.c
@@ -78,7 +78,9 @@ struct mtev_http_connection {
 struct mtev_http_request {
   struct bchain *first_input; /* The start of the input chain */
   struct bchain *last_input;  /* The end of the input chain */
-  struct bchain *current_input;  /* The point of the input where we */
+  struct bchain *current_input;  /* The point of the input where we are */
+  struct bchain *user_data;  /* user consumption reads from this which holds dechunked data */
+  struct bchain *user_data_last;  /* end of the user_data chain */
   size_t         current_offset; /* analyzing. */
   mtev_boolean freed;
 
@@ -1118,7 +1120,8 @@ mtev_http_ctx_acceptor_free(void *v) {
   mtev_http_ctx_session_release((mtev_http_session_ctx *)v);
 }
 static int
-mtev_http_session_req_consume_chunked(mtev_http_session_ctx *ctx,
+mtev_http_session_req_consume_chunked(mtev_http_session_ctx *ctx, 
+                                      mtev_compress_type compression_type,
                                       int *mask) {
   /* chunked encoding read */
   int next_chunk = -1;
@@ -1147,6 +1150,7 @@ mtev_http_session_req_consume_chunked(mtev_http_session_ctx *ctx,
     if(str_in_f) {
       unsigned int clen = 0;
       const char *cp = in->buff + in->start;
+      const char *cp_begin = in->buff + in->start;
       while(cp <= str_in_f) {
         if(*cp >= '0' && *cp <= '9') clen = (clen << 4) | (*cp - '0');
         else if(*cp >= 'a' && *cp <= 'f') clen = (clen << 4) | (*cp - 'a' + 10);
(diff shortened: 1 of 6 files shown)


rileyberton:
Tested with python, ruby, nodejs (via snowth test suite), C (libcurl), and command line curl.  I think this is correct now.

Note that ruby doesn't appear to handle the gzip format we return without writing some code: (see autodetect gzip header)

```header = {'Content-Type' => 'text/json'}
request = Net::HTTP::Post.new(uri.request_uri, header)
request["Accept-Encoding"] = "gzip"
request.body = params.to_json

response = http.request(request)
puts "Response body is: #{response.body.length}"

# auto detect gzip header
zi = Zlib::Inflate.new(Zlib::MAX_WBITS + 32)

uncompressed = zi.inflate(response.body)```

File: src/mtev_http.c [hunk 1]
```diff
@@ -1118,66 +1120,88 @@ mtev_http_ctx_acceptor_free(void *v) {
...
+              goto successful_chunk_size;
+            } else {
+              /**
+               * we have decoded a chunk length but the current bchain
+               * is not large enough to handle the entire chunk.
+               * 
+               * In this case we allocate a new bchain which is large
+               * enough to hold the entire chunk, copy in the chunk data that
+               * we have already read and then keep reading.
+               */
```
aidancully: I think this is allocating when we haven't filled in the required size - the better check would be to allocate only when the required size > in->allocd? With the code in its current form, I think in the common case that we've received enough to see the chunk size, but haven't received the entire chunk yet, we'll basically do a pointless copy of the previously read data into a new chunk before trying to read again? And we'll do that until we get a full read?

src/mtev_http.c
```diff
@@ -1197,20 +1221,57 @@ mtev_http_session_req_consume_chunked(mtev_http_session_ctx *ctx,
...
+      }
+      memcpy(data->buff + data->size, in->buff + in->start, MIN(in->size, next_chunk));
+      data->size = MIN(in->size, next_chunk);
+      in->start += data->size;
+      in->size -= data->size;
+      ctx->req.user_data_last = data;
+      if (ctx->req.payload_chunked) {
+        /* there must be a \r\n at the end of this block */
+        str_in_f = strnstrn("\r\n", 2, in->buff + in->start, in->size);
+        if(str_in_f != (in->buff + in->start)) {
```
aidancully: I _think_ this can just be a `if (in->size < 2 || strcmp(in->buff + in->start, "\r\n") != 0) {`? `in->start` should now point to the end of the chunk, so it seems a little more correct, to me... more robust against possibly storing extra data in the `in` chunk.

src/mtev_http.c
```diff
@@ -1118,66 +1120,88 @@ mtev_http_ctx_acceptor_free(void *v) {
...
+        unsigned int clen = 0;
+        const char *cp = in->buff + in->start;
+        const char *cp_begin = in->buff + in->start;
+        while(cp <= str_in_f) {
+          if(*cp >= '0' && *cp <= '9') clen = (clen << 4) | (*cp - '0');
+          else if(*cp >= 'a' && *cp <= 'f') clen = (clen << 4) | (*cp - 'a' + 10);
+          else if(*cp >= 'A' && *cp <= 'F') clen = (clen << 4) | (*cp - 'A' + 10);
+          else if(*cp == '\r' && cp[1] == '\n') {
+            mtevL(http_debug, "Found for chunk length(%d)\n", clen);
+            next_chunk = clen;
```
aidancully: I think `next_chunk = clen` should only be set in the body of the `if` below?
```
if (in->size - 2 >= clen) {
  next_chunk = clen;
  ...
```

src/mtev_http.c
rileyberton: No, `in->size` here represents unread raw (still chunk encoded) data that we read off the wire.  If there isn't enough bytes left in the raw block, we move the bytes from the raw block to a new block which *is* large enough to hold the chunk size.

aidancully: yes, but AIUI in->size is the wrong thing to check if the block is big enough. in->allocd is the size of the block, in->size is what's been read so far. Only need to allocate a new block in in->allocd is too small?

rileyberton: we aren't checking the size of `in`, we are checking to see if `in` contains the entire chunk.  If it does not since there isn't enough `in->size` left, we make new block that starts at `in->size` but when we read we increase the `new_in->size` so that clen will be satisfied.

The only danger here is extremely slow reads which will allocate and free unneccessarily.. But since the default block size is 32K and the CURL uses 16K chunks, it would have to be an impossibly slow read.

aidancully: To be clear, i'm not saying that the code is wrong, but that it will allocate and free unnecessarily. Which is possible due to some combination of slow reads and large chunks... Large chunks are the bigger concern (e.g. 1MB size chunk with read returning 16K each time would memcpy 1MB (2^20) / 16KB (2^14) == 64 (2^6) times, with each subsequent time copying an additional 16K of data, such that the penultimate time through this loop we've copied (1MB - 64KB) of data needlessly? Admittedly, I don't know how common such large chunks would be for our workloads... But I think it's ok to just do the allocation / copy once, on actually determining what the size of the chunk is?

aidancully: that is, I think this would work: instead of:
```
} else {
```
use
```
} else if (in->allocd < clen) {
```

rileyberton: I don't think that's true.  If the chunk size (`clen`) was 1MB we would make a 1MB block and read into it.  Only in the case where we had really slow reads would we waste allocations and frees.  If the read after the 1MB alloc did not read 1MB we would do this again.

That code block is not correct as `in->allocd` will always be larger than `clen` for small chunk sizes (32K allocd vs 16K chunk size) and we would never allocate the replacement block.

It might work if you did:

```} else if ((in->allocd - in->start) < clen) { ```

But I don't think we need to worry about this for now.

aidancully: Agree that it's not very consequential, I didn't expect this to get so drawn out, I'm willing to drop the point.

aidancully:
LGTM

PR #198, opened 2017-02-09
First pass at getting FreeBSD builds

postwait:


src/utils/mtev_time.c
```diff
@@ -228,14 +228,23 @@ fetch_cclock_scale(int cpuid, struct cclock_scale *cs) {
 }
 #endif
 
-#if defined(linux) || defined(__linux) || defined(__linux__)
+#if defined(BSD) || defined(__FreeBSD__)
```
esproul: Note that MacOS also defines `BSD`, so maybe this needs to go after the `__MACH__` case?

esproul:
`ctfconvert` on FreeBSD 10 throws this error for every object:

    ERROR: ctfconvert: failed to initialize DWARF: Wrong CU version. Only 2 and 3 supported [dwarf_init_info(538)]

and at the end:

    ERROR: ctfmerge: No ctf sections found to merge

If we want to support FreeBSD 10, and it's not hard to fix, I suggest we do it. Otherwise, we can make the minimum version FreeBSD 11.

esproul:
The ctfconvert errors on 10 are not fatal. We end up not having CTF in the shared object, but we don't fail to build.  It seems like the CTF tools on FreeBSD 10 are just not capable of dealing with the DWARF version 4 symbols emitted by gcc.  The CTF tools on FreeBSD 11 seem fine.

PR #201, opened 2017-02-11
Fix leaks

postwait:


postwait:
This is ::findleaks clean now.

PR #203, opened 2017-02-12
Base add/mod/dev in epoll off e->mask.

postwait:
We should trust that if e->mask is zero, then the event should not already
be registered with epoll.  If it is zero, we should not DEL and ADD in lieu
of MOD.  If it is non-zero, we should DEL and MOD in lieu of ADD.


rileyberton:
LGTM

PR #205, opened 2017-02-14
Fix typo

jabley:
Spefically -> Specifically

Not sure if this needs changing in the HTML as well? I'm assuming that
is built from the Markdown, but I've not looked at the build system yet.


postwait:
The HTML will get committed when I next do a docs build.

`gmake docs-html` in the top-level.

PR #206, opened 2017-02-15
fix weird gzip encoding issue when output buffer is not large enough …

rileyberton:
…to process input buffer and there is data to flush


pamaddox:
LGTM

PR #207, opened 2017-02-15
Rework stack traces.

postwait:
Refactor setup and calling code, separate into separate source file.
mtev_main will now setup sefl-diagnosis for foreground operated processes.


pamaddox:
LGTM

PR #213, opened 2017-02-20
add mtev_now_ms and mtev_now_us, add docs for mtev_time.h

rileyberton:
Added convenience functions for getting the current timeofday (wall time) in milliseconds and microseconds.  Also updated docs for mtev_time.h to be our doxygen compatible.


docs-md/apireference/c.md
```diff
@@ -979,6 +993,22 @@ mtev_getip_ipv4(struct in_addr remote, struct in_addr *local)
...
+mtev_gettimeofday(struct timeval *t, void **ttp)
+```
+
+ *  
+> Maybe fast-pathed version of gettimeofday
+
+
+ *    * **RETURN** same as system gettimeofday();
+ * 
+ * If the fast path is taken, tzp is ignored.
```
esproul: is it `tzp` or `ttp`, as in the prototype?

docs-md/apireference/c.md
```diff
@@ -1307,6 +1367,116 @@ mtev_sort_set_next_function(void *current, void *value)
...
+```
+
+ *  
+> Usually this is managed for you, but this is safe to call at any time
+
+
+ *    * **RETURN** mtev_true if it was successful in parameterizing the CPU for rdtsc, mtev_false otherwise
+ * 
+ * safe to call at any time but if you start_tsc, you should never need to call this
+ * as the maintenance system can do it for you.. However, if you find you need to call it
```
esproul: Double period after "you"

docs-md/apireference/c.md
```diff
@@ -1307,6 +1367,116 @@ mtev_sort_set_next_function(void *current, void *value)
...
+```
+
+ *  
+> will switch on/off rdtsc usage across all cores regardless of detected state of rdtsc or start/stop usage.
+
+
+ * 
+ * Defaults to enabled.
+ * 
+ * This is idependent of start_tsc/stop_tsc.  You can disable all and reenable and the thread
```
esproul: "independent"

rileyberton:
LGTM from slack chat.

PR #217, opened 2017-03-02
AMQP module via librabbitmq

postwait:


src/modules/mtev_amqp.c
```diff
@@ -0,0 +1,566 @@
...
+   will send outbound data in tight bunches every 50ms.
+
+   This function does work until it can't make progress (error) and then
+   returns to the caller (which will shutdown the connection and attempt to
+   bring it back online).
+ */
+static void
+handleconn(struct amqp_conn *conn) {
+  while(1) {
+    struct timeval small_sleep = { 0, 100000 }; /* 50ms */
```
rileyberton: `static struct timeval small_sleep = { 0, 10000 };` and hoist out of the loop?

src/modules/mtev_amqp.c
```diff
@@ -0,0 +1,566 @@
...
+
+    amqp_maybe_release_buffers(conn->conn);
+
+    amqp_rpc_reply_t ret = amqp_consume_message(conn->conn, &envelope, &small_sleep, 0);
+
+    if (ret.reply_type == AMQP_RESPONSE_NORMAL) {
+      amqp_envelope_t *env = malloc(sizeof(*env));
+      memcpy(env, &envelope, sizeof(*env));
+      ck_fifo_spsc_enqueue_lock(&conn->inbound);
+      ck_fifo_spsc_entry_t *fifo_entry = malloc(sizeof(*fifo_entry));
```
rileyberton: I think `ck_fifo_spsc` queue requires recycling items:

    ck_fifo_spsc_entry_t *fifo_entry = ck_fifo_spsc_recycle(&conn->inbound);
    if (fifo_entry == NULL) {
      fifo_entry = malloc(sizeof(*fifo_entry));
    }

Otherwise the dequeue leaks.

src/modules/mtev_amqp.c
```diff
@@ -0,0 +1,566 @@
...
+    }
+    eventer_set_fd_blocking(amqp_socket_get_sockfd(socket));
+    int status = amqp_socket_open_noblock(socket, conn->host, conn->port, &fiveseconds);
+    if (status) {
+      mtevL(nlerr, "AMQP error connecting socket: %s\n", strerror(errno));
+      goto teardown;
+    }
+
+    die_on_amqp_error(
+      amqp_login(conn->conn, conn->vhost, 0, 131072, 0, AMQP_SASL_METHOD_PLAIN,
```
rileyberton: what is 131072?  Some buffer size?

src/modules/mtev_amqp.c
```diff
@@ -0,0 +1,566 @@
...
+    env = (void *)((uintptr_t)env | mask);
+    copies[connection_id_broadcast_if_negative] = env;
+  }
+
+  for (i=0; i<the_conf->number_of_conns; i++) {
+    struct amqp_conn *cc = &the_conf->amqp_conns[i];
+    if(connection_id_broadcast_if_negative < 0 ||
+       i == connection_id_broadcast_if_negative) {
+      ck_fifo_spsc_enqueue_lock(&cc->outbound);
+      ck_fifo_spsc_entry_t *fifo_entry = malloc(sizeof(*fifo_entry));
```
rileyberton: Same as earlier.  Have to recycle.

src/modules/mtev_amqp.c
postwait: Well... the param to consume message isn't const, so I worry it might update it.  I always want it to be 1s, so it seemed safest to explicitly ensure that before each call.

src/modules/mtev_amqp.c
postwait: max frame size. from the amqp example code.

rileyberton:
Looks good to me

PR #218, opened 2017-03-02
Mtev dyn buffer

rileyberton:


src/utils/mtev_dyn_buffer.h
```diff
@@ -0,0 +1,137 @@
...
+ */
+typedef struct mtev_dyn_buffer {
+  uint8_t static_buffer[4096];
+  size_t size;
+  uint8_t *data;
+  uint8_t *pos;
+} mtev_dyn_buffer_t;
+
+/*! \fn void mtev_dyn_buffer_init(mtev_dyn_buffer_t *buf)
+ *  \brief initialize a dyn_buffer
```
postwait: I *think* the docs will not generate correctly with block *'s  It should be white-space up to the \ commands on all likes but the opening comment.

rileyberton: Yes, you are right, adds the extra '*' to docs.  Will fix.

rileyberton: docs fixed

PR #219, opened 2017-03-10
Implement eventer_add_asynch_dep(...)

postwait:
This new function can be called from an asynchronous job to
delay the completion of the current executing job until the
new added deap has past its cleanup phase.

### DIFF
--- a/src/eventer/eventer_impl.c
+++ b/src/eventer/eventer_impl.c
@@ -777,7 +777,29 @@ void eventer_add_asynch(eventer_jobq_t *q, eventer_t e) {
     job->timeout_event->callback = eventer_jobq_execute_timeout;
     eventer_add(job->timeout_event);
   }
-  eventer_jobq_enqueue(q ? q : __default_jobq, job);
+  eventer_jobq_enqueue(q ? q : __default_jobq, job, NULL);
+}
+
+void eventer_add_asynch_dep(eventer_jobq_t *q, eventer_t e) {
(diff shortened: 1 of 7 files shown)


aidancully:
I've got an alternative approach for this use case, I'll try to PR that, and we can compare... I think it wouldn't be hard to create a fork/join implementation that can be used with jobq's (like this does), or from other threads..

postwait:
it would be nice to get some action on this.. I've brought this up to current master.

src/eventer/eventer_impl.c
```diff
@@ -777,7 +777,29 @@ void eventer_add_asynch(eventer_jobq_t *q, eventer_t e) {
...
+  if(e->whence.tv_sec) {
+    job->timeout_event = eventer_alloc();
+    job->timeout_event->thr_owner = e->thr_owner;
+    memcpy(&job->timeout_event->whence, &e->whence, sizeof(e->whence));
+    job->timeout_event->mask = EVENTER_TIMER;
+    job->timeout_event->closure = job;
+    job->timeout_event->callback = eventer_jobq_execute_timeout;
+    eventer_add(job->timeout_event);
+  }
+  eventer_jobq_enqueue(q ? q : __default_jobq, job, eventer_jobq_inflight());
```
rileyberton: `eventer_jobq_enqueue(job->jobq, ...);` eliminates the second branch here.

rileyberton:
Looks good to me

PR #220, opened 2017-03-14
Enable SSL for mtev_websocket_client.c

gallison93:
Based on `mtev_connection_complete_connect` and `mtev_connection_ssl_upgrade` from `mtev_reverse_socket.c`.

Enabling SSL breaks connections to non-SSL endpoints: `eventer_SSL_connect` returns `EAGAIN` so we return the given mask and are never triggered again. I'm not sure how the eventer SSL code is (not?) supposed to handle a non-SSL connection, so I provided an alternate code path to avoid the SSL upgrade and opset switch by not providing an `sslconfig`.

I'm also at a loss as to how the `sslconfig` should be acquired, and would appreciate feedback on that. For now I've simply stuck in a new `mtev_hash_table` parameter for `mtev_websocket_client_new`.

That prototype is getting a bit lengthy now. I'm intending to make another PR this week to fix an outstanding bug/design problem in this file(that prevents freeing clients/causes invalid memory accesses and undefined behavior) that may require another parameter in `_new`, so I'll very likely add variants(e.g. `_new` would do no SSL on default pool, `_new_pool` accepts a pool, `_new_pool_ssl` accepts pool and sslconfig, rough concept, will refine it) then rather than in this PR.

### DIFF
--- a/src/mtev_websocket_client.c
+++ b/src/mtev_websocket_client.c
@@ -28,9 +28,11 @@ struct mtev_websocket_client {
   pthread_mutex_t lock;
   const char *path;
   const char *host;
+  int port;
   const char *service; /* protocol */
   int wanted_eventer_mask;
   char client_key[25];
+  mtev_hash_table *sslconfig;
   void *closure;
 };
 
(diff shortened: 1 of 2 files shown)


rileyberton:
Looks good to me

PR #222, opened 2017-03-20
Make Linux epoll impl more resilient to confused add/del/mod.

postwait:
This change isn't ideal as I think it accommodates other bugs, but
it would appear that we can remove and fd and then re-trigger it
with a mask and an event in the master_fds.  This causes us to
attempt to add when we should modify (or mod when we should add).
This simply provides an explicit fallback if the approach we take
errors b/c it was backwards, we try the other operation.

### DIFF
--- a/src/eventer/eventer_epoll_impl.c
+++ b/src/eventer/eventer_epoll_impl.c
@@ -153,6 +153,7 @@ static void eventer_epoll_impl_add(eventer_t e) {
   lockstate = acquire_master_fd(e->fd);
   master_fds[e->fd].e = e;
 
+  mtevL(eventer_deb, "epoll_ctl(%d, add, %d)\n", spec->epoll_fd, e->fd);
   rv = epoll_ctl(spec->epoll_fd, EPOLL_CTL_ADD, e->fd, &_ev);
   if(rv != 0) {
     mtevFatal(mtev_error, "epoll_ctl(%d,add,%d,%x) -> %d (%d: %s)\n",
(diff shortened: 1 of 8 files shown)


test/pool-shift-async/test.c
```diff
@@ -0,0 +1,124 @@
...
+  eventer_t conne = mtev_http_connection_event_float(mtev_http_session_connection(restc->http_ctx));
+  if(conne) eventer_remove_fd(conne->fd);
+
+  eventer_t e = eventer_alloc();
+  e->callback = doasynchstuff;
+  e->closure = restc;
+  e->mask = EVENTER_ASYNCH;
+  eventer_add(e);
+
+usleep(1000);
```
aidancully: is this intentional? `usleep` in the event loop? if so, it may deserve a comment...

src/eventer/eventer_epoll_impl.c
```diff
@@ -314,26 +319,41 @@ static void eventer_epoll_impl_trigger(eventer_t e, int mask) {
...
+        int epoll_rv;
+        int epoll_cmd = (e->mask == 0 || needs_add) ? EPOLL_CTL_ADD : EPOLL_CTL_MOD;
         spec = eventer_get_spec_for_event(e);
-        if(epoll_ctl(spec->epoll_fd, epoll_cmd, fd, &_ev) != 0) {
+        mtevL(eventer_deb, "epoll_ctl(%d, %s, %d)\n", spec->epoll_fd, epoll_cmd == EPOLL_CTL_ADD ? "add" : "mod", fd);
+        epoll_rv = epoll_ctl(spec->epoll_fd, epoll_cmd, fd, &_ev);
+        if(epoll_rv != 0 &&
+           ((epoll_cmd == EPOLL_CTL_ADD && errno == EEXIST) ||
+            (epoll_cmd == EPOLL_CTL_MOD && errno == ENOENT))) {
+            /* try the other way */
```
aidancully: seems like this should have an `mtevL(eventer_deb` when we do this.

aidancully:
Can't say I completely understand the case being addressed, but at the least nothing seems harmful, and I believe the change makes us more robust. LGTM.

PR #225, opened 2017-03-23
prevent race condition between wslay_event_send and wslay_event_queue_msg

gallison93:
This change uses existing locks to make calls to `wslay_event_send` and `wslay_event_queue_msg` mutually exclusive.

Wslay in general does not appear thread-safe. Specific to our usage of it, `wslay_event_send` and `wslay_event_queue_msg` both manipulate the same internal queue in a given `wslay_event_context_ptr`. There are several potential failures from this, although in testing I only saw one: if a push(in `wslay_event_queue_msg`) and pop(in `wslay_event_send`) operation are interleaved in a certain manner, wslay will interpret the queue as empty from that point forward and thus stop writing frames even as more and more messages are successfully enqueue'd to send(which according to wslay docs should cause wslay to attempt more writes). There is no error provided, neither to our network peer nor in return values from wslay calls, nor is the `wslay_event_context_ptr` considered closed or write-disabled.

### DIFF
--- a/src/mtev_http.c
+++ b/src/mtev_http.c
@@ -1608,7 +1608,6 @@ wslay_send_callback(wslay_event_context_ptr ctx,
   mtev_http_session_ctx *session_ctx = user_data;
   session_ctx->wanted_eventer_mask = 0;
 
-  pthread_mutex_lock(&session_ctx->write_lock);
   if(!session_ctx->conn.e || session_ctx->is_websocket == mtev_false) {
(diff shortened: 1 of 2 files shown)


gallison93:
`mtev_websocket_client_drive` should be restricted to a single thread for a given client, so I believe that lock is safe from recursion. `mtev_websocket_client_send` could theoretically be called by any number of threads at any time, but it makes no recursive call nor does it offer a callback so I believe that should also be free from recursion.

PR #226, opened 2017-03-24
Fix log initialization startup sequence.

postwait:
During log init, we create the run-time dynamic instances
of "stderr", "debug", "error", and "notice," but previously
they didn't match their boot equivalents.  Fixing this allows
us to default the log creation state to enabled.

### DIFF
--- a/src/utils/mtev_log.c
+++ b/src/utils/mtev_log.c
@@ -66,6 +66,9 @@
 #define LIBMTEV_LOG_ENABLED() 0
 #endif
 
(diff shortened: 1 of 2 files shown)


src/utils/mtev_log.c
```diff
@@ -401,19 +408,51 @@ static void materialize_deps(mtev_log_stream_t ls) {
...
+  }
+}
+
+static void
+mtev_log_materialize() {
+  mtev_hash_iter iter = MTEV_HASH_ITER_ZERO;
+
+  while(mtev_hash_adv(&mtev_loggers, &iter)) {
+    mtev_log_stream_t ls = iter.value.ptr;
+    debug_printf("dematerializing(%s)\n", ls->name);
```
rileyberton: debug should say "materializing"

PR #230, opened 2017-03-27
Add OmniOS buildtools

HeinrichHartmann:
Allow for to control libmtev build from a remote omnios host
Add librabbitmq dependency

### DIFF
--- a/buildtools/omnios/configure.sh
+++ b/buildtools/omnios/configure.sh
@@ -1,0 +1,11 @@
+autoreconf -i
+
+./configure \
+  --prefix=/opt/circonus/ \
+  --exec-prefix=/opt/circonus \
+  --sysconfdir=/opt/circonus/etc \
+  --includedir=/opt/circonus/include \
+  --bindir=/opt/circonus/bin \
+  --sbindir=/opt/circonus/sbin \
+  --libdir=/opt/circonus/lib/amd64/ \
+  --libexecdir=/opt/circonus/libexec/amd64/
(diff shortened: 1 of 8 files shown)


buildtools/omnios/configure.sh
```diff
@@ -0,0 +1,11 @@
+autoreconf -i
+
+./configure \
+  --prefix=/opt/circonus/ \
+  --exec-prefix=/opt/circonus \
+  --sysconfdir=/opt/circonus/etc \
+  --includedir=/opt/circonus/include \
```
postwait: Shouldn't this be /opt/circonus/include/amd64 ?

HeinrichHartmann: I copied this over from snowh. But I can change it if needed.

HeinrichHartmann:
This is a circonus-labs repo, so I don't see anything wrong with us using `/opt/criconus` paths for our development.

If we are going to support Vagrant / Docker builds in the future (I think we should) build information is needed inside this repo. Packaging could then use these scrpts as well.

Even if we don't use them elsewhere and they run out of date, then they are more likely to be working than copy/paste from the readme.

HeinrichHartmann:
Using --prefix=/usr/local breaks mtevbusted since it expects luamtev in /opt/circonus:
https://github.com/circonus-labs/libmtev/blob/master/test/mtevbusted-script.in#L1

Maybe that should be fixed.

postwait:
That's a bug in mtevbusted.. it should respect the install prefix.

postwait:
Circonus specific build stuff should live in a different tree.

PR #235, opened 2017-03-29
Add C++ demangling to the stacktrace output.

postwait:
Pulled an isolated demangler from the Google Android project.
https://android.googlesource.com/platform/external/gcc-demangle/

Added some hacky features to make it not allocate.


aidancully:
LGTM - works on both Linux and illumos

PR #239, opened 2017-04-01
add mtev_websocket_client_new_noref

gallison93:
This fulfills the use case where a mtev_websocket_client_t is solely interacted with through the callbacks.

The example cleanup handler in `src/examples/websocket_client.c` was deceptive: it suggested that one could safely `mtev_websocket_client_free()` the client in the cleanup handler. This was not true, and that actually caused unsafe memory accesses after the handler returned. The example was so small that the issue would almost certainly never appear there.

In the use case mentioned above, one would either be forced to leak the clients or track references to clients just to free them later.


src/mtev_websocket_client.c
```diff
@@ -397,10 +398,14 @@ mtev_websocket_client_complete_connect(eventer_t e, int mask, void *closure, str
...
 #endif
 
-mtev_websocket_client_t *
-mtev_websocket_client_new(const char *host, int port, const char *path, const char *service,
-                          mtev_websocket_client_callbacks *callbacks, void *closure, eventer_pool_t *pool,
-                          mtev_hash_table *sslconfig) {
+// using an extra static function rather than just calling `_new` from
+// `_new_noref` and then setting the field because it is possible(but
+// unlikely), if the eventer_t was added to a different thread, for the client
+// to have fired and failed and cleaned up incorrectly before the field is set
```
postwait: C comments please /* */

src/mtev_websocket_client.h
```diff
@@ -32,6 +32,22 @@ API_EXPORT(mtev_websocket_client_t *)
...
+    2. The caller must not let a reference to the client escape from the provided callbacks.
+*/
+API_EXPORT(mtev_boolean)
+  mtev_websocket_client_new_noref(const char *url, int port,
+                                  const char *path, const char *service,
+                                  mtev_websocket_client_callbacks *callbacks,
+                                  void *closure,
+                                  eventer_pool_t *pool,
+                                  mtev_hash_table *sslconfig);
+
```
postwait: While you're in here writing all this commentary -- please write it in doxygen /*! \fn style (see other headers as example).  This will auto-produce documentation.

gallison93: I've added doxygen comments for this header. How does auto-producing the documentation work?

src/mtev_websocket_client.h
```diff
@@ -24,49 +24,148 @@ typedef struct {
...
                                              mtev_websocket_client_cleanup_callback cleanup_callback);
 
+/*! \fn mtev_boolean mtev_websocket_client_send(mtev_websocket_client_t *client, int opcode, void *buf, size_t len)
+    \brief Enqueue a message
+    \param client
+    \param opcode opcode as defined in RFC 6455 and referenced in wslay.h
+    \param buf pointer to buffer containing data to send
+    \param len number of bytes of buf to send
+    \return boolean indicating success/failure
+    This function makes a copy of buf of length len.
```
postwait: Need a blank line between \return and follow-on documentation.. .(here and everywhere)

src/mtev_websocket_client.h
```diff
@@ -24,49 +24,153 @@ typedef struct {
...
+    3. Out of memory.
+*/
 API_EXPORT(mtev_boolean)
   mtev_websocket_client_send(mtev_websocket_client_t *client, int opcode,
                              void *buf, size_t len);
 
+/*! \fn void mtev_websocket_client_free(mtev_websocket_client_t *client)
+    \brief Free a client
+    \param client client to be freed
+    This function will cleanup the client(and hence trigger any set cleanup_callback) first.
```
postwait: Sorry to be unclear the first time.  The block of description must be separated from the  slash commands (or the text will be wrapped up into the last command).

src/mtev_websocket_client.h
postwait:
```
gmake docs-serve
```

is what I use.  Ultimately we run `gmake docs-html` and commit any new/changes so that it goes live at http://circonus-labs.github.io/libmtev/

I try to make sure that the docs-md changes are part of my commits for features, but the commits to changes in docs/ (from `gmake docs-html`) are a separate commit as they are big and noisy.

PR #242, opened 2017-04-07
LZ4F support on the response side

rileyberton:
* add a curl helper you can hand to CURLOPT_WRITEFUNCTION that can decompress LZ4F responses
* fix LZ4F handling so it actually works on the response side

### DIFF
--- a/src/utils/mtev_compress.c
+++ b/src/utils/mtev_compress.c
@@ -13,6 +13,7 @@ struct mtev_stream_compress_ctx
 {
   mtev_compress_type type;
   mtev_boolean begun;
+  mtev_boolean flushed;
   LZ4F_compressionContext_t lz4_compress_ctx;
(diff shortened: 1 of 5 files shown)


src/utils/mtev_compress.c
```diff
@@ -24,12 +25,19 @@ struct mtev_stream_decompress_ctx
...
+  mtev_stream_decompress_ctx_t decompress_ctx;
+};
+
 size_t
 mtev_compress_bound(mtev_compress_type type, size_t source_len)
 {
   switch(type) {
   case MTEV_COMPRESS_LZ4F:
-    return LZ4F_compressBound(source_len, NULL);
+    return LZ4F_compressBound(source_len, NULL) + 15;
```
postwait: 15? Seems magical? #define?

postwait: Or a comment?

PR #244, opened 2017-04-13
fix potential use-after-free in eventer cross-thread triggers.

aidancully:
CTT's _first_ add the eventer_t to a cross-thread-trigger list,
_then_ call eventer_wakeup(event). If the event is handled across
threads before the wakeup, then the event might be free'd before
eventer_wakeup has a chance to fully examine it. Protect against
this by refcnt'ing the eventer_t around the transaction.


rileyberton:
Looks good to me

PR #245, opened 2017-04-14
Handle triggers of events from within event handlers on the same thre…

rileyberton:
…ad.  Treat these the same as a cross thread trigger

### DIFF
--- a/src/eventer/eventer_epoll_impl.c
+++ b/src/eventer/eventer_epoll_impl.c
@@ -218,10 +218,21 @@ static void eventer_epoll_impl_update(eventer_t e, int mask) {
     if(e->mask & EVENTER_READ) _ev.events |= (EPOLLIN|EPOLLPRI);
     if(e->mask & EVENTER_WRITE) _ev.events |= (EPOLLOUT);
     if(e->mask & EVENTER_EXCEPTION) _ev.events |= (EPOLLERR|EPOLLHUP);
-    mtevL(eventer_deb, "epoll_ctl(%d, mod, %d)\n", spec->epoll_fd, e->fd);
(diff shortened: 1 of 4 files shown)


src/eventer/eventer_epoll_impl.c
```diff
@@ -285,9 +296,24 @@ static void eventer_epoll_impl_trigger(eventer_t e, int mask) {
...
+    mtevL(eventer_deb, "Incoming event: %p already owned by this thread\n", e);
+    /* The incoming triggered event is already owned by this thread.  
+       This means our floated event completed before the current
+       event handler even exited.  So it retriggered recursively
+       from inside the event handler.  
+       
+       Treat this special case the same as a cross thread trigger
+       and just queue this event to be picked up on the next loop
+    */
+    eventer_cross_thread_trigger(e, mask);
```
aidancully: This approach will lead to this log:
```
    if(master_fds[fd].e != NULL) {
      mtevL(eventer_deb, "Attempting to trigger already-registered event fd: %d cross thread.\n", fd);
    }
```
because we've already set `master_fds[fd].e = e;` before calling cross_thread_trigger, so when we re-enter this handler cross-thread, `master_fds[fd].e != NULL`, and we hit that debug log. Given the (commented-out) `mtevAssert` in that code block, this seems contrary to intent.

What about changing the `if(master_fds[fd].e == NULL)` case above like follows:
```c
if(master_fds[fd].e == NULL) {
  lockstate = acquire_master_fd(fd);
  if (lockstate == EV_ALREADY_OWNED) {
    eventer_cross_thread_trigger(e,mask);
    return;
  }
  master_fds[fd].e = e;
  e->mask = 0;
  needs_add = 1;
}
```
?

aidancully:
LGTM

PR #246, opened 2017-04-21
Mtev logbuf

aidancully:


postwait:
I'm curious as to what problem this solves. It appears to only delay the probe effect instead of avoid it.

The logging subsystem should already elide blocks when the log is disabled or has no output.  That's the purpose of the materialization of the log_stream DAG.  Under what circumstances are we seeing issues that this would address?

PR #247, opened 2017-04-21
speculative mtev_log facility

aidancully:


src/utils/mtev_log.h
```diff
@@ -108,6 +108,10 @@ API_EXPORT(mtev_log_stream_t)
   mtev_log_stream_new_on_fd(const char *, int, mtev_hash_table *);
 API_EXPORT(mtev_log_stream_t)
   mtev_log_stream_new_on_file(const char *, mtev_hash_table *);
+API_EXPORT(mtev_log_stream_t) mtev_log_speculate(int nlogs, int nbytes);
+API_EXPORT(void)
+  mtev_log_speculate_finish(mtev_log_stream_t ls, mtev_log_stream_t speculation);
```
postwait: The ls being null is unclear.  However... perhaps another argument that takes an enum for commit or rollback... or

```
#define MTEV_LOG_SPECULATE_ROLLBACK ((mtev_log_stream_t)NULL)
```

PR #248, opened 2017-04-27
NULL out chains passed to RELEASE_BCHAIN

rileyberton:


src/mtev_http.c
```diff
@@ -1272,7 +1275,6 @@ mtev_http_session_req_consume_read(mtev_http_session_ctx *ctx,
       data->size = MIN(in->size, next_chunk);
       in->start += data->size;
       in->size -= data->size;
-      ctx->req.user_data_last = data;
```
rileyberton: Because it was already set a few lines above.

postwait:
Lgtm

PR #250, opened 2017-05-02
Fix epoll

rileyberton:


aidancully:
LGTM

src/eventer/eventer_epoll_impl.c
```diff
@@ -287,31 +287,51 @@ static void eventer_epoll_impl_trigger(eventer_t e, int mask) {
...
+       *
+       * Treat this special case the same as a cross thread trigger
+       * and just queue this event to be picked up on the next loop
+       */
       eventer_cross_thread_trigger(e, mask);
       return;
     }
-    release_master_fd(fd, lockstate);
+    /*
+     * If we are readding the event to the master list here, also do the needful
```
postwait: needful?

src/eventer/eventer_epoll_impl.c
```diff
@@ -287,31 +287,51 @@ static void eventer_epoll_impl_trigger(eventer_t e, int mask) {
...
+    struct epoll_event _ev;
+    memset(&_ev, 0, sizeof(_ev));
+    _ev.data.fd = fd;
+    spec = eventer_get_spec_for_event(e);
+    if(mask & EVENTER_READ) _ev.events |= (EPOLLIN|EPOLLPRI);
+    if(mask & EVENTER_WRITE) _ev.events |= (EPOLLOUT);
+    if(mask & EVENTER_EXCEPTION) _ev.events |= (EPOLLERR|EPOLLHUP);
+
+    mtevL(eventer_deb, "epoll_ctl(%d, add, %d)\n", spec->epoll_fd, fd);
+    if (epoll_ctl(spec->epoll_fd, EPOLL_CTL_ADD, fd, &_ev) != 0) {
```
postwait: What if `mask & (EVENTER_{READ,WRITE,EXCEPTION}) == 0`... we'll attempt to add "nothing".

rileyberton: can mask be zero here?  We are triggered from something.. maybe an assertion that mask != 0?

src/eventer/eventer_epoll_impl.c
rileyberton: https://en.wikipedia.org/wiki/Do_the_needful

PR #253, opened 2017-05-15
use locks to protect against cross thread releases

rileyberton:


rileyberton:
snowth ci passes with these changes

PR #255, opened 2017-05-17
add mtev_lru

rileyberton:
Simple LRU with configurable free function for evicted items.

Went with `ck_hs` directly instead of `mtev_hash_table` because I needed access to `ck_hs_set` to catch conflicting puts.

### DIFF
--- a/src/utils/mtev_lru.c
+++ b/src/utils/mtev_lru.c
@@ -1,6 +1,7 @@
 #include "mtev_lru.h"
 #include "mtev_hash.h"
 #include "mtev_log.h"
+#include "mtev_rand.h"
 
 #include <ck_hs.h>
 #include <sys/queue.h>
@@ -43,8 +44,6 @@ static struct ck_malloc malloc_ck_hs = {
   .free = lru_free
 };
 
-static int rand_init = 0;
-
 static unsigned long
 lru_entry_hash(const void *k, unsigned long seed)
 {
@@ -57,13 +56,10 @@ lru_entry_hash(const void *k, unsigned long seed)
 static void
 hs_init(ck_hs_t *hs, unsigned int mode, ck_hs_hash_cb_t *hf, ck_hs_compare_cb_t *cf, unsigned long size)
 {
-  if(!rand_init) {
-    srand48((long int)time(NULL));
-    rand_init = 1;
-  }
+  mtev_rand_init();
   struct ck_malloc *allocator = &malloc_ck_hs;
 
-  if (ck_hs_init(hs, mode | CK_HS_MODE_SPMC, hf, cf, allocator, size, lrand48()) == false) {
+  if (ck_hs_init(hs, mode | CK_HS_MODE_SPMC, hf, cf, allocator, size, mtev_rand()) == false) {
     mtevFatal(mtev_error, "Cannot initialize ck_hs\n");
   }
 }
--- a/src/utils/mtev_lru.c
+++ b/src/utils/mtev_lru.c
@@ -188,6 +188,7 @@ mtev_lru_put(mtev_lru_t *lru, const char *key, size_t key_len, void *val)
   }
   if (previous) {
     struct lru_entry *p = container_of(previous, struct lru_entry, key);
+    TAILQ_REMOVE(&lru->lru_cache, p, list_entry);
     lru->free_fn(p->entry);
     free(p);
   }
--- a/src/utils/mtev_lru.c
+++ b/src/utils/mtev_lru.c
@@ -231,7 +231,7 @@ mtev_lru_remove(mtev_lru_t *c, const char *key, size_t key_len)
     free(r);
 
     /* when removing, perform GC every 100K rows */
-    if (c->lru_cache_size % GC_CADENCE == 0) {
+    if (ck_hs_count(&c->hash) > 0 && c->lru_cache_size % GC_CADENCE == 0) {
       ck_hs_gc(&c->hash, GC_CADENCE, (rand() % ck_hs_count(&c->hash)));
     }
   }
--- a/src/utils/mtev_lru.c
+++ b/src/utils/mtev_lru.c
@@ -4,6 +4,7 @@
 
 #include <ck_hs.h>
 #include <sys/queue.h>
+#include <stddef.h>
 
 struct lru_entry {
   void *entry;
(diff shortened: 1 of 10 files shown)


src/utils/mtev_lru.c
```diff
@@ -0,0 +1,252 @@
+#include "mtev_lru.h"
+#include "mtev_hash.h"
+#include "mtev_log.h"
+
+#include <ck_hs.h>
+#include <sys/queue.h>
```
aidancully: I _think_ sys/queue.h exists on all the platforms we care about (not sure about Windows?), but it makes me nervous to use it, since it's not POSIX.

src/utils/mtev_lru.c
```diff
@@ -0,0 +1,252 @@
...
+  ((size_t)(&((type *)0)->member))
+
+#define container_of(derived_ptr, type, field)                \
+  ((type *)((char *)(derived_ptr) - offset_of(type, field)))
+
+static void
+hs_init(ck_hs_t *hs, unsigned int mode, ck_hs_hash_cb_t *hf, ck_hs_compare_cb_t *cf, unsigned long size)
+{
+  if(!rand_init) {
+    srand48((long int)time(NULL));
```
aidancully: note that this can potentially re-seed the client application's lrand48 PRNG, which seems undesirable.

src/utils/mtev_lru.c
rileyberton: exists on omni and centos and ubuntu.

src/utils/mtev_lru.c
```diff
@@ -0,0 +1,252 @@
...
+static int rand_init = 0;
+
+#define HS_HF_STRING(F)                         \
+  static unsigned long                          \
+  F(const void *k, unsigned long seed)          \
+  {                                             \
+    return mtev_hash__hash(k, strlen(k), seed);   \
+  }
+
+HS_HF_STRING(lru_entry_hash);
```
rileyberton: no good reason, this was repurposed from some personal code I wrote a while ago for experiments.  Will replace.

src/utils/mtev_lru.c
```diff
@@ -0,0 +1,252 @@
...
+
+#define offset_of(type, member)                 \
+  ((size_t)(&((type *)0)->member))
```
rileyberton: I wasn't sure if this would exist on all platforms.. I know that `container_of` does not

src/utils/mtev_lru.c
rileyberton: don't have a good answer here.  We do the same thing in `mtev_hash_init`

aidancully: OK... then we should have this code use the same rand-init that `mtev_hash_init` does, maybe extract that to another function that this can call? I don't really understand how libck uses that random number, but it makes it feel a little worse that this will make `mtev_hash_init` and `mtev_lru` will probably get the same output from `lrand48`...

src/utils/mtev_lru.c
aidancully: yes, `container_of` does not... but `offsetof` is part of C89.

src/utils/mtev_lru.c
```diff
@@ -0,0 +1,252 @@
...
+void
+mtev_lru_invalidate(mtev_lru_t *lru)
+{
+  pthread_mutex_lock(&lru->mutex);
+  void *entry;
+  ck_hs_iterator_t it = CK_HS_ITERATOR_INITIALIZER;
+  while(ck_hs_next(&lru->hash, &it, &entry) == true) {
+    struct lru_entry *e = container_of(entry, struct lru_entry, key);
+    lru->free_fn(e->entry);
+    free(e);
```
aidancully: looks like this will `free` the entry for the `ck_hs`, but we haven't actually updated the `ck_hs` to remove the entry? this is fine when `mtev_lru_invalidate` is called by `mtev_lru_destroy`, but not fine when `mtev_lru_invalidate` is called otherwise...

aidancully: I think it's possible to use the `lru->lru_cache` to walk this list, freeing as you go... Should be more efficient (better locality of reference), allow you to empty the `ck_hs` in one fell swoop before you start, and re-initialize the `lru_cache` at the end when it gets fully emptied. Then this function could be usable by clients in its own right...

src/utils/mtev_lru.c
rileyberton: added `mtev_rand` to address this and replaced all `lrand48()` with `mtev_rand()`

src/utils/mtev_lru.c
```diff
@@ -0,0 +1,252 @@
...
+    TAILQ_INSERT_HEAD(&c->lru_cache, e, list_entry);
+  }
+  c->lru_cache_size++;
+}
+
+static void 
+touch_lru_cache_no_lock(mtev_lru_t *c, struct lru_entry *e)
+{
+  if (c->max_entries > 0) {
+    if (e->list_entry.tqe_next != NULL || e->list_entry.tqe_prev != NULL) {
```
aidancully: I assume this function relies on the LRU entry being on the lru_cache, in which case you don't need to move the entry if it's already at the head. And it's probably better to use the TAILQ macros to peek into the structure, than to rely on their implementation details?
```
if (TAILQ_PREV(e, list_entry) != NULL) {
  TAILQ_REMOVE(&c->lru_cache, e, list_entry);
  TAILQ_INSERT_HEAD(&c->lru_cache, e, list_entry);
}
```

src/utils/mtev_lru.c
```diff
@@ -0,0 +1,252 @@
...
+  pthread_mutex_lock(&lru->mutex);
+  void *previous = NULL;
+  if (ck_hs_set(&lru->hash, hash, e->key, &previous) == false) {
+    free(e);
+    pthread_mutex_unlock(&lru->mutex);
+    return mtev_false;
+  }
+  if (previous) {
+    struct lru_entry *p = container_of(previous, struct lru_entry, key);
+    lru->free_fn(p->entry);
```
aidancully: unlink `p` from `lru->lru_cache`?

src/utils/mtev_lru.c
rileyberton: yes, that was the intention.. missing `ck_hs_reset`

rileyberton:
I think that's all of the problems.

src/utils/mtev_lru.c
```diff
@@ -231,7 +231,7 @@ mtev_lru_remove(mtev_lru_t *c, const char *key, size_t key_len)
     free(r);
 
     /* when removing, perform GC every 100K rows */
-    if (c->lru_cache_size % GC_CADENCE == 0) {
+    if (ck_hs_count(&c->hash) > 0 && c->lru_cache_size % GC_CADENCE == 0) {
```
aidancully: how fast is `ck_hs_count`? maybe use `! TAILQ_EMPTY(&c->lru_cache)` instead of calling that function?

src/utils/mtev_lru.c
```diff
@@ -230,10 +230,13 @@ mtev_lru_remove(mtev_lru_t *c, const char *key, size_t key_len)
     rval = r->entry;
     free(r);
 
-    /* when removing, perform GC every 100K rows */
-    if (ck_hs_count(&c->hash) > 0 && c->lru_cache_size % GC_CADENCE == 0) {
+    /* when removing, perform GC every GC_CADENCE changes */
+    int ec = ck_pr_load_32(&c->expire_count);
+    if (ec == GC_CADENCE && ck_hs_count(&c->hash) > 0) {
```
aidancully: same comment... seems inefficient to count the number of elements in the cache when all you care about is non-empty, for which you can use `TAILQ_EMPTY`.

aidancully:
a couple of nits, but nothing that should stop the merge.

src/utils/mtev_lru.c
rileyberton: ultra fast.. it keeps an internal count and just returns that number

aidancully: ok

src/utils/mtev_lru.c
rileyberton: they are the same speed and we are dealing with the hashset here.  the count check is just to avoid the divide by zero error

aidancully: :+1:

PR #256, opened 2017-05-18
Add mtev_uuid_copy Inline Function

pamaddox:
libuuid's uuid_copy function copies character-by-character. This is
pretty inefficient. Add a mtev_uuid_copy function that just does a
memcpy.

Also - add a mtev_uuid.h header to allow just including that to get all
of the mtev uuid utilities.


rileyberton:
Looks good to me

PR #259, opened 2017-05-25
Fix Memory Leak In mtev_http_complete_request

pamaddox:
If we try to eventer_read and get back 0, don't immediately go to
full_error.... try to finalize, then go to full_error if that fails.
This will clean up the memory properly, which wasn't always being done
before.


rileyberton:
Looks good to me

PR #260, opened 2017-05-26
Realloc Request And Response At Next Req

pamaddox:
When we hit the "next_req" flag, we need to realloc the request and
response. Otherwise, we could end up allocating data into a structure
flagged "freed", leading to memory leaks.


rileyberton:
Looks good to me

PR #263, opened 2017-06-08
allow naming of threads within the eventer

postwait:


src/eventer/eventer_impl.c
```diff
@@ -98,8 +98,16 @@ struct eventer_impl_data {
 };
 
 static __thread struct eventer_impl_data *my_impl_data;
+static __thread char *thread_name;
 static struct eventer_impl_data *eventer_impl_tls_data = NULL;
 
+void eventer_set_thread_name(const char *name) {
+  thread_name = strdup(name);
```
pamaddox: Is it possible to destroy a thread? If so - does this clean up?

pamaddox:
Changes look good to me.

PR #264, opened 2017-06-12
implement hooks in lua

postwait:


src/modules/lua.c
```diff
@@ -73,11 +73,17 @@ mtev_lua_lmc_alloc(mtev_dso_generic_t *self, mtev_lua_resume_t resume) {
 void
 mtev_lua_lmc_free(lua_module_closure_t *lmc) {
   if(lmc) {
+    if(lmc->lua_state) lua_close(lmc->lua_state);
```
aidancully: shouldn't the `mtev_lua_states_lock` surround the `lua_close`?

postwait: The lmc is thread local.. that state will only ever be seen/manipulated by a single thread. so I think we're safe to do this locklessly.

aidancully: there are a couple of functions that peek into the lua_state via `mtev_lua_states`: `mtev_console_lua_thread_reporter_json`, `mtev_console_lua_thread_reporter_ncct`. I was worried they'd try looking into a `lua_state` that got closed.

aidancully: In that case, why do you need the lock at all? AFAICT, it exists solely to protect `mtev_lua_states` from access from multiple threads, but if everything's in one thread, that'd be unnecessary? Or I may have miscommunicated... I'm not worried about multiple threads talking to `lmc`, but about multiple threads talking to `lmc->lua_state`, which is accessible from other threads via that hash-table.

postwait: probably don't.  I imagine they existed when mtev_hash_table's weren't thread-safe.  I was just conforming to using the locks to protect the hash table from concurrent manipulation as is done elsewhere.  I can look at removing the locks as unnecessary in a subsequent changeset.

aidancully:
LGTM

PR #265, opened 2017-06-14
Add convenience functions for resuming after flaot

postwait:


src/mtev_http.c
```diff
@@ -404,6 +404,14 @@ eventer_t mtev_http_connection_event_float(mtev_http_connection *conn) {
   }
   return e;
 }
+void mtev_http_connection_resume_after_float(mtev_http_connection *conn) {
+  if(conn->e) {
```
rileyberton: Don't we have to check for `NULL` here?

rileyberton:
Looks good to me

PR #266, opened 2017-06-14
Add `fanout`, `fanout_pool`, and `accept_thread` listener attributes.

postwait:
You can now, via configuration, make an listener have a dedicated
accept thread as well as indicate that new connections should be
fanned out over arbitrary pools.


rileyberton:
Looks good to me

PR #270, opened 2017-06-16
Add event contexts and a zipkin layer for them.

postwait:
* Allow external components to register eventer contexts that
   provide contextural data and ghosting of eventer creation,
   copying, deallocation, and callback invocation.

 * Provide a TLS that represents the event currently in callback.

 * Add a zipkin event context layer and connect it to the HTTP
   server such that inbould requests attach spans to the connection
   if a span exists and have that span persisted forward through
   causally created events.

 * Provide a zipkin event context client span managment API and
   provide mtev_zipkin_curl.h as a sample client wrapper.

### DIFF
--- a/src/eventer/eventer.c
+++ b/src/eventer/eventer.c
@@ -50,6 +50,46 @@ stats_handle_t *eventer_unnamed_callback_latency;
 static mtev_atomic64_t ealloccnt;
 static mtev_atomic64_t ealloctotal;
 
+static struct {
+  char *name;
+  eventer_context_opset_t *opset;
+} eventer_contexts[MAX_EVENT_CTXS];
+static int eventer_contexts_cnt = 0;
+
+void eventer_callback_prep(eventer_t e, int m, void *c, struct timeval *n) {
+  for(int i=0; i<eventer_contexts_cnt; i++) {
+    if(eventer_contexts[i].opset->eventer_t_callback_prep) {
+      eventer_contexts[i].opset->eventer_t_callback_prep(e,m,c,n);
+    }
+  }
+}
+
+void eventer_callback_cleanup(eventer_t e, int m) {
+  for(int i=0; i<eventer_contexts_cnt; i++) {
+    if(eventer_contexts[i].opset->eventer_t_callback_cleanup) {
+      eventer_contexts[i].opset->eventer_t_callback_cleanup(e,m);
+    }
+  }
+}
+
+int eventer_register_context(const char *name, eventer_context_opset_t *o) {
+  int idx = eventer_contexts_cnt++;
+  eventer_contexts[idx].name = strdup(name);
+  eventer_contexts[idx].opset = o;
+  return idx;
+}
+void *eventer_get_context(eventer_t e, int idx) {
+  if(idx < 0 || idx >= eventer_contexts_cnt) return NULL;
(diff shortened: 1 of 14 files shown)


src/eventer/eventer_impl_private.h
```diff
@@ -144,6 +152,20 @@ void eventer_impl_init_globals(void);
 void eventer_dispatch_recurrent(void);
 void eventer_dispatch_timed(struct timeval *next);
 void eventer_mark_callback_time(void);
+void eventer_set_this_event(eventer_t e);
+void eventer_callback_prep(eventer_t, int, void *, struct timeval *);
+void eventer_callback_cleanup(eventer_t, int);
+
+static inline int RUN_CALLBACK(eventer_t e, int m, void *c, struct timeval *n) {
```
rileyberton: I thought this was a macro when reading higher up.. maybe `eventer_run_callback` so it's clear it's a function?

src/mtev_rest.c
```diff
@@ -447,6 +454,8 @@ mtev_http_rest_new_rule_auth_closure(const char *method, const char *base,
   }
   rule = calloc(1, sizeof(*rule));
   rule->method = strdup(method);
+  rule->nice_name = malloc(strlen(base) + strlen(expr) + 2);
```
rileyberton: I don't think we care but this name leaks?

src/utils/mtev_zipkin_curl.h
```diff
@@ -37,11 +37,42 @@
 
 static inline struct curl_slist *
 mtev_zipkin_inst_curl_headers(struct curl_slist *inheaders) {
+  static const char *curl_name = "curl request";
```
rileyberton: "curl request to: $URL"?

src/eventer/eventer.c
```diff
@@ -50,6 +50,46 @@ stats_handle_t *eventer_unnamed_callback_latency;
...
+  for(int i=0; i<eventer_contexts_cnt; i++) {
+    if(eventer_contexts[i].opset->eventer_t_callback_cleanup) {
+      eventer_contexts[i].opset->eventer_t_callback_cleanup(e,m);
+    }
+  }
+}
+
+int eventer_register_context(const char *name, eventer_context_opset_t *o) {
+  int idx = eventer_contexts_cnt++;
+  eventer_contexts[idx].name = strdup(name);
```
rileyberton: I think this name leaks.

src/mtev_rest.c
postwait: We don't have a way to remove (or free) rules that have been added... it will need to be address with the freeing of everything related to that rule if/when we add that capability.

src/utils/mtev_zipkin_curl.h
postwait: Technically we don't have the URL yet when that call is made.  But the URL is already passed as a binary annotation.  I could provide an alternative form of that function for use with the naming of curl requests arbitrarily.

src/eventer/eventer.c
postwait: it's global, write-once.  Not leaked; owned.

src/eventer/eventer_impl_private.h
postwait: Makes sense. I'll make that change.

PR #271, opened 2017-06-19
Fix lua gc of events.

postwait:
The lua event cleanup had an inverted assumption. When an event goes
"live" in the schedule such that lua is off-cpu, we should zero out
the eventer pointer because if the state is "cleaned up" the event
will be freed in due course of the eventer callbacks and lua is not
responsible for GCing it.

Make the coro state hashes safer by copying the key we use.


aidancully:
LGTM

PR #273, opened 2017-06-26
Mtev waterlevel2

aidancully:


aidancully:
I'm changing this a little, making the levels unsigned, fixing a bug in when we'd signal ENABLE. (The argument that negative values sometimes make sense applied to the other approach, but does not apply here.)

src/utils/mtev_flow_regulator.h
```diff
@@ -0,0 +1,62 @@
...
+typedef struct _mtev_flow_regulator_t mtev_flow_regulator_t;
+
+typedef enum {
+  MTEV_FLOW_REGULATOR_TOGGLE_DISABLED=-2,
+  MTEV_FLOW_REGULATOR_TOGGLE_DISABLE=-1,
+  MTEV_FLOW_REGULATOR_TOGGLE_KEEP=0,
+  MTEV_FLOW_REGULATOR_TOGGLE_ENABLE=1,
+} mtev_flow_regulator_toggle_t;
+
+mtev_flow_regulator_t *mtev_flow_regulator_create(unsigned int low, unsigned int high);
```
postwait: Please add doxygen comments so that this will be added into the general MTEV docs

PR #280, opened 2017-07-11
Flow regulator stabilize

aidancully:
remove `mtev_flow_regulator_stabilize` API, replace with `stable_try_raise_one` and `stable_lower`.

### DIFF
--- a/test/flowreg/test.conf
+++ b/test/flowreg/test.conf
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf8" standalone="yes"?>
-<flowreg_test lockfile="/tmp/example.lock">
+<flowreg_test>
   <!--
   <watchdog glider="/opt/local/bin/bt"
             tracedir="/var/cores/example"/>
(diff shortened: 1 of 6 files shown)


test/flowreg/test.conf
```diff
@@ -0,0 +1,63 @@
+<?xml version="1.0" encoding="utf8" standalone="yes"?>
+<flowreg_test lockfile="/tmp/example.lock">
```
postwait: I'd remove the lockfile.

test/flowreg/test.conf
```diff
@@ -0,0 +1,63 @@
...
+      <debug>
+        <outlet name="debug"/>
+        <log name="debug/eventer" disabled="true"/>
+        <log name="debug/example" disabled="true"/>
+      </debug>
+    </components>
+  </logs>
+  <modules directory="/path/to/modules">
+  </modules>
+  <listeners>
```
postwait: I'd remove all the listeners as well.

PR #288, opened 2017-08-03
docs/development/mtev_main.html: fix typo

nshalman:


nshalman:
That makes much more sense.

PR #290, opened 2017-08-07
Implement mtev_conf_register_global_param

postwait:


src/mtev_conf.h
```diff
@@ -44,6 +44,26 @@
...
+  MTEV_PARAM_DOUBLE,
+  MTEV_PARAM_BOOLEAN
+} mtev_param_type_t;
+
+typedef mtev_boolean (*mtev_param_validator_t)(mtev_param_type_t, void *mem);
+mtev_boolean
+mtev_conf_register_global_param(const char *name, const char *xpath,
+                                mtev_param_type_t ptype, void *mem,
+                                mtev_param_validator_t vf);
+
```
aidancully: I'd consider using a parser / display function, rather than a type and validator? e.g.,
```c
typedef int (*mtev_param_parser_t)(const struct _mtev_param *param, const char *arg);
typedef struct _mtev_param {
  const mtev_param_parser_t *parser;
  const char *help_str;
  void *data;
} mtev_param_t
```
? This _could_ be useful for allowing more semantic information in the parser, e.g., to be able to parse `1day12hr`, or perhaps enum variants, and keeps the validation with the parsing... mtev would then define standard `mtev_param_t`'s for things like booleans...

I also think it'd be neat for this work with the `mtev_confstr` facility for parsing.

src/mtev_conf.c
```diff
@@ -94,6 +96,114 @@ static xmlDocPtr master_config = NULL;
...
+  p->name = strdup(name);
+  if(xpath) p->xpath = strdup(xpath);
+  p->ptype = ptype;
+  p->memory = mem;
+  p->validate = validate;
+  if(mtev_hash_store(&global_param_sets, p->name, strlen(p->name), p))
+    return mtev_true;
+  free(p->name);
+  free(p->xpath);
+  free(p);
```
aidancully: should this try to fetch the data from `mtev_conf`? to enforce that the memory agress with the config file on start-up.

postwait: I worry that it might be too much change as code various places has interesting defaulting logic.  I was going for plumbing not fixtures: a way to update params online.

src/mtev_conf.c
```diff
@@ -94,6 +96,114 @@ static xmlDocPtr master_config = NULL;
...
+        if(running) *running = mtev_true;
+      }
+      else if(!strcasecmp(value, "off") || !strcasecmp(value, "false")) {
+        b = mtev_true;
+        if(p->validate && !p->validate(p->ptype, &b)) return mtev_true;
+        *((mtev_boolean *)p->memory) = b;
+        if(running) *running = mtev_true;
+      }
+      else {
+        return mtev_true;
```
postwait: The function is supposed to return false if the parameter name is not found; true otherwise.

The two other params state whether the actions were successful.

src/mtev_conf.c
aidancully: OK..

src/mtev_conf.h
postwait: I'd like to toss the confstr stuff.  It's used inconsistently and didn't take into consideration the complexities that the current config system handles.

The idea of a parser sounds fairly useful, but a parse has to look at text which makes the common case of simply validating a parsed value much harder and more repetitive on the caller.  So if we add a parser, it should be in addition to the validator -- which only validates the actual value.

src/mtev_conf.c
aidancully: then the calling code (line 3081 as I write this) needs to be updated to observe the other params on "success"..

postwait: +1 fixed.

aidancully:
LGTM

postwait:
Added a parser

aidancully:
LGTM

PR #293, opened 2017-08-09
Add thread-safety to mtev_conf.

postwait:
This is an API/ABI breaking change.  Efforts were made to
preserve allocation model and be as load as possible on old
use.  One specific note is the need to release section(s)
fetched via mtev_conf_get_section(s).

### DIFF
--- a/src/modules/lua_mtev.c
+++ b/src/modules/lua_mtev.c
@@ -2695,16 +2695,19 @@ nl_conf_get_string(lua_State *L) {
     char *element, *base;
     SPLIT_PATH(path, base, element);
     
-    section = mtev_conf_get_section(NULL, base);
-    if(!section || !element) lua_pushboolean(L, 0);
+    section = mtev_conf_get_section(MTEV_CONF_ROOT, base);
+    if(mtev_conf_section_is_empty(section) || !element) {
+      lua_pushboolean(L, 0);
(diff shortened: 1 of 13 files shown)


src/modules/lua_mtev.c
```diff
@@ -2720,7 +2723,7 @@ nl_conf_get_string_list(lua_State *L) {
   const char *base_path = lua_tostring(L,1);
   const char *child_path = lua_tostring(L,2);
 
-  mtev_conf_section_t* mqs = mtev_conf_get_sections(NULL, base_path, &cnt);
+  mtev_conf_section_t* mqs = mtev_conf_get_sections(MTEV_CONF_ROOT, base_path, &cnt);
```
gallison93: looks like we could leave due to an error on line 2736 without calling `mtev_conf_release_sections()`

src/modules/mtev_amqp.c
```diff
@@ -398,7 +398,7 @@ init_conns(void) {
       goto bail;
     }
   }
-  free(mqs);
+  mtev_conf_release_sections(mqs, the_conf->number_of_conns);
```
gallison93: the goto on line 398 skips this `mtev_conf_release_sections()`. line 405 probably needs to be updated

PR #296, opened 2017-08-10
Coverity fixes

SamuelWM01:
Misc fixes that were required in coverity. Most are removals of dead code. Functionality appears to have not been interrupted as gmake built correctly and all tests were successful.

### DIFF
--- a/src/noitedit/read.c
+++ b/src/noitedit/read.c
@@ -271,8 +271,6 @@ read_char(EditLine *el, char *cp)
 		else
 			num_read = read(el->el_infd, cp, 1);
 
-		if (num_read)
-			break;
 		if (num_read == -1) {
 			if (errno == EAGAIN) return 0;
 		 	if (!tried && read__fixio(el->el_infd, errno) == 0) {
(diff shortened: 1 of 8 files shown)


src/noitedit/read.c
```diff
@@ -271,8 +271,6 @@ read_char(EditLine *el, char *cp)
 		else
 			num_read = read(el->el_infd, cp, 1);
 
-		if (num_read)
-			break;
```
postwait: This is unlikely to be correct.  I think it should say and be changed to `if(num_read > 0) break;`

src/utils/mtev_security.c
```diff
@@ -135,7 +135,10 @@ mtev_security_usergroup(const char *user, const char *group, mtev_boolean effect
           pwnam_buflen = safe_size * 2;
           safe_size = MAX(grnam_buflen, pwnam_buflen);
           free(mallocd);
-          buf = mallocd = malloc(safe_size);
+          if(safe_size >= 0)
```
postwait: Change this to an: `mtevAssert(safe_size > 0);` and remove the else and `printf` -- we can't use printf here as we may not be connected to standard output.

SamuelWM01:
noticed a typo fixing

PR #298, opened 2017-08-10
LFU cache with O(1) operations based on bucketed lists of frequencies

rileyberton:
Need this to deal with currently unbounded caches in a way that's smarter than LRU.

### DIFF
--- a/src/utils/mtev_lfu.c
+++ b/src/utils/mtev_lfu.c
@@ -135,13 +135,12 @@ mtev_lfu_invalidate(mtev_lfu_t *lfu)
   pthread_mutex_unlock(&lfu->mutex);
 }
 
-/* returns the frequency of the list it was in */
 static void
 remove_from_frequency_list_no_lock(mtev_lfu_t *lfu, struct lfu_entry *le)
 {
   STAILQ_REMOVE(&le->frequency_list_head->lfu_cache, le, lfu_entry, freq_list_entry);
   if (STAILQ_FIRST(&le->frequency_list_head->lfu_cache) == NULL) {
-    
+
     /* we can remove the entire frequency bucket */
     STAILQ_REMOVE(&lfu->lfu_frequency_list, le->frequency_list_head, lfu_cache_entry, list_entry);
     free(le->frequency_list_head);
@@ -149,8 +148,8 @@ remove_from_frequency_list_no_lock(mtev_lfu_t *lfu, struct lfu_entry *le)
   le->frequency_list_head = NULL;
 }
 
-/* this requires a lock to be held elsewhere 
- * 
+/* this requires a lock to be held elsewhere
+ *
  * There are potentially several items in the same
  * frequency bucket, we remove the first one
  */
@@ -161,7 +160,7 @@ expire_least_lfu_cache_no_lock(mtev_lfu_t *c)
   struct lfu_cache_entry *e = STAILQ_FIRST(&c->lfu_frequency_list);
   struct lfu_entry *le = STAILQ_FIRST(&e->lfu_cache);
 
-  (void)remove_from_frequency_list_no_lock(c, le);
+  remove_from_frequency_list_no_lock(c, le);
 
   c->lfu_cache_size--;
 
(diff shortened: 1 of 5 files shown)


src/utils/mtev_lfu.c
```diff
@@ -0,0 +1,319 @@
...
+    }
+    free(e);
+  }
+
+  ck_hs_reset(&lfu->hash);
+  lfu->lfu_cache_size = 0;
+  pthread_mutex_unlock(&lfu->mutex);
+}
+
+/* returns the frequency of the list it was in */
```
aidancully: nit: this comment can't be correct... returns `void`.

src/utils/mtev_lfu.c
```diff
@@ -0,0 +1,319 @@
...
+  ck_hs_remove(&c->hash, hash, le->key);
+
+  int ec = ck_pr_load_32(&c->expire_count);
+  if (ec == GC_CADENCE) {
+    c->expire_count = 0;
+    ck_hs_gc(&c->hash, GC_CADENCE, (rand() % ck_hs_count(&c->hash)));
+  }
+  c->expire_count++;
+
+  if (le->entry != NULL) {
```
aidancully: not sure I agree with only calling the free_fn when the entry is non-null... Possible use for this would be to store, say, file-descriptors in a hash, and avoid allocations by storing them as `(void *) (uintptr_t) fd`. `fd == 0` is valid, and we'd want to be able to close it if the free-function invokes `close(2)`.

src/utils/mtev_lfu.c
```diff
@@ -0,0 +1,319 @@
...
+    }
+    /* there is a 1 bucket.  add to tail */
+    le->frequency_list_head = e;
+    STAILQ_INSERT_TAIL(&e->lfu_cache, le, freq_list_entry);
+  }
+  c->lfu_cache_size++;
+}
+
+static struct lfu_cache_entry *
+new_frequency_bucket(mtev_lfu_t *c, size_t freq, struct lfu_cache_entry *prior)
```
aidancully: lock probably needs to be held, here, so rename `_no_lock`?

src/utils/mtev_lfu.c
```diff
@@ -0,0 +1,319 @@
...
+add_lfu_cache_no_lock(mtev_lfu_t *c, struct lfu_entry *le)
+{
+  if (c->max_entries > 0) {
+    if ((c->lfu_cache_size + 1) > c->max_entries) {
+      expire_least_lfu_cache_no_lock(c);
+    }
+    /* new entries always go in the list at frequency == 1 */
+    struct lfu_cache_entry *e = STAILQ_FIRST(&c->lfu_frequency_list);
+    if (e == NULL || e->frequency > 1) {
+      /* make a frequency bucket of 1 */
```
aidancully: for code re-use, should call `new_frequency_bucket(_no_lock)`

src/utils/mtev_lfu.c
```diff
@@ -0,0 +1,319 @@
...
+      e = malloc(sizeof(struct lfu_cache_entry));
+      STAILQ_INIT(&e->lfu_cache);
+      e->frequency = 1;
+      STAILQ_INSERT_HEAD(&c->lfu_frequency_list, e, list_entry);
+    }
+    /* there is a 1 bucket.  add to tail */
```
aidancully: missing `else {`?

aidancully:
this doesn't need to be addressed now, or possibly ever, but i'm also a little worried about the number of allocations this may result in for the frequency buckets. unless i'm missing something, it seems likely that common uses of this facility will end up with a frequency bucket per item, and that the buckets will often not be adjacent... with current implementation, this could mean an alloc and a free every time we update the frequency of an item? potentially every time we fetch it from the LFU?

I don't think that's a showstopper for the merge - we'll see if that's an issue later, under load, and if so, we can fix the implementation w/o breaking the ABI.

src/utils/mtev_lfu.c
aidancully: never mind, this is correct as written.

src/utils/mtev_lfu.c
rileyberton: yes, sorry, changed api halfway through, forgot to remove comment.

src/utils/mtev_lfu.c
rileyberton: ok, makes sense

rileyberton:
Not sure the allocations are super problematic.  They might become so in cases where there is really high churn in the LFU.  I think in practice, frequently used items will get higher frequency counts and those buckets will exist.  

There might be an issue with the `1` bucket under high churn.  We could introduce a freelist here to combat that.

aidancully:
For example, if we have a single element in the LFU, we call `mtev_lfu_get` to retrieve it, which calls `touch_lfu_get_no_lock`. `touch_lfu_get_no_lock` will remove the element from its existing bucket (to bump the frequency), freeing the old bucket (there was only one element), and allocating a new bucket for the new bucket. This is potentially a `free` / `alloc` on every fetch.

rileyberton:
Yes but that condition (1 element in the bucket) would only be under very low usage of the LFU.  We could retain the bucket once it's created but this has the same unbounded memory problem.

aidancully:
I suspect that 1 element in the bucket will be the common case, as the LFU's lifetime increases. In other words, I think it will, after running for a long time, be _unlikely_ that frequently-used elements will have the exact same (or +/- 1) use frequencies. Say, after running for a long time, we ultimately fetch an item 1,000,000 times... I think it will be unlikely that any other elements will have been accessed exactly 999,999, or 1,000,001 times.

Again, though, I don't think it blocks the merge.

rileyberton:
So the other way to handle this is to use a skiplist or something for cache entries and pay logN to search it for insertion every time there is an operation (we'd have to remove from the list and re-add on frequency change to ensure the order is correct).

We could also implement a min-heap or other ordered tree and pay the same cost.

aidancully:
The 1 element bucket pathology can be largely avoided by not immediately freeing the bucket after removal, and instead using the old structure if there isn't already a destination bucket. This would have the effect of, essentially, bumping the frequency field of the `lfu_cache_entry` for the common case that the `lfu_cache_entry` has only one element, and no adjacent neighbors.

PR #299, opened 2017-08-11
Misc coverity fixes

SamuelWM01:
Made changes to many null check deferences.
I don't know to much about this bug type so a solid review would be appreciated.
Also made a few other coverity fixes.


SamuelWM01:
making corrections to incorrect fixes.

PR #300, opened 2017-08-14
re-use LFU buckets instead of allocating, make the zero sized LFU wor…

rileyberton:
…k properly

### DIFF
--- a/src/utils/mtev_lfu.c
+++ b/src/utils/mtev_lfu.c
@@ -135,17 +135,24 @@ mtev_lfu_invalidate(mtev_lfu_t *lfu)
   pthread_mutex_unlock(&lfu->mutex);
 }
 
-static void
+static struct lfu_cache_entry *
 remove_from_frequency_list_no_lock(mtev_lfu_t *lfu, struct lfu_entry *le)
 {
+  struct lfu_cache_entry *rval = NULL;
   STAILQ_REMOVE(&le->frequency_list_head->lfu_cache, le, lfu_entry, freq_list_entry);
   if (STAILQ_FIRST(&le->frequency_list_head->lfu_cache) == NULL) {
 
     /* we can remove the entire frequency bucket */
     STAILQ_REMOVE(&lfu->lfu_frequency_list, le->frequency_list_head, lfu_cache_entry, list_entry);
-    free(le->frequency_list_head);
+    /* don't free, instead return the bucket for possible reuse */
+    rval = le->frequency_list_head;
+    rval->list_entry.stqe_next = NULL;
+    rval->frequency = 0;
+    rval->lfu_cache.stqh_first = NULL;
+    rval->lfu_cache.stqh_last = &rval->lfu_cache.stqh_first;
   }
   le->frequency_list_head = NULL;
+  return rval;
 }
 
 /* this requires a lock to be held elsewhere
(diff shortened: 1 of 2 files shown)


src/utils/mtev_lfu.c
```diff
@@ -220,12 +236,23 @@ touch_lfu_cache_no_lock(mtev_lfu_t *c, struct lfu_entry *e)
...
         next_bucket == STAILQ_FIRST(&c->lfu_frequency_list) ||
         next_bucket->frequency != bucket->frequency + 1) {
-      next_bucket = new_frequency_bucket_no_lock(c, bucket->frequency + 1, bucket);
+      need_new_bucket = mtev_true;
+    }
+    struct lfu_cache_entry *empty = remove_from_frequency_list_no_lock(c, e);
+    if (need_new_bucket) {
+      if (empty != NULL) {
+        next_bucket = empty;
+        next_bucket->frequency = bucket->frequency + 1;
```
aidancully: Just reading this, but isn't `next_bucket` == `bucket`? and wasn't `bucket->frequency` zero'd out `remove_from_frequency_list_no_lock`? need to cache the frequency, I think, not re-use it from the bucket.

src/utils/mtev_lfu.c
```diff
@@ -236,6 +263,12 @@ touch_lfu_cache_no_lock(mtev_lfu_t *c, struct lfu_entry *e)
 mtev_boolean
 mtev_lfu_put(mtev_lfu_t *lfu, const char *key, size_t key_len, void *val)
 {
+  /* a max size of zero means to disable the LFU */
+  if (lfu->max_entries == 0) {
+    /* we don't want calling code to error */
+    return mtev_true;
```
aidancully: Why don't you want calling code to error, here? The way this is written, this signals success, but `mtev_lfu_get` and `mtev_lfu_remove` will not return a valid value...

src/utils/mtev_lfu.c
rileyberton: Yes, good catch.

src/utils/mtev_lfu.c
rileyberton: `mtev_lfu_get` or `remove` could always return an invalid value because something else may have come along to evict that item at any time.

What I wanted to avoid was someone writing code like:

```
if (mtev_lfu_put(lfu, mykey, mykeylen, myval) == mtev_false) {
  mtevFatal(mtev_error, "Cannot cache item in LFU\n");
}
```
And then changing the LFU size to zero in the config so they can disable it causing the app to crash.

aidancully: Whenever an item gets evicted, the free callback gets invoked. (Aside: might be better to rename this the eviction callback?) Most clients, I think, will have the idea that if an item is successfully inserted into the LFU, it will be possible to retrieve it from the LFU unless it was evicted. A reasonable alternative implementation would be to immediately invoke the free callback if you try to `mtev_lfu_put` and the max size is 0?

rileyberton: The `free_fn` case is a good point.  I will have it return false.  I think the auto-free functionality would be unexpected.

src/utils/mtev_lfu.c
```diff
@@ -266,7 +266,7 @@ mtev_lfu_put(mtev_lfu_t *lfu, const char *key, size_t key_len, void *val)
   /* a max size of zero means to disable the LFU */
   if (lfu->max_entries == 0) {
     /* we don't want calling code to error */
```
aidancully: remove comment

aidancully:
LGTM

PR #301, opened 2017-08-14
fix race in eventer_update_whence

postwait:


src/eventer/eventer_impl.c
```diff
@@ -854,13 +854,14 @@ eventer_t eventer_remove_timed(eventer_t e) {
   pthread_mutex_unlock(&t->te_lock);
   return removed;
 }
-void eventer_update_timed(eventer_t e, int mask) {
+void eventer_update_timed(eventer_t e, int mask, struct timeval *new_whence) {
```
aidancully: can we change the name of this function? with the existing name, if any external code _used_ to call `eventer_update_timed`, it'd break now that the signature is different... I think it'd be better to break at dynamic link time than at run-time?

postwait: good call.

aidancully:
LGTM

PR #302, opened 2017-08-15
more tests for edge cases and fix edge cases

rileyberton:
Found a few edge cases:

1) getting the same value twice from the LFU would segfault
2) there is a case where buckets would get ordered incorrectly

### DIFF
--- a/src/utils/mtev_lfu.c
+++ b/src/utils/mtev_lfu.c
@@ -11,7 +11,7 @@
 
 struct lfu_cache_entry;
 
-STAILQ_HEAD(frequency_list, lfu_cache_entry);
+TAILQ_HEAD(frequency_list, lfu_cache_entry);
 
 struct lfu_entry {
   void *entry;
@@ -24,7 +24,7 @@ struct lfu_entry {
 struct lfu_cache_entry {
   size_t frequency;
   STAILQ_HEAD(lfu_cache, lfu_entry) lfu_cache;
-  STAILQ_ENTRY(lfu_cache_entry) list_entry;
+  TAILQ_ENTRY(lfu_cache_entry) list_entry;
 };
 
 struct mtev_lfu {
@@ -92,7 +92,7 @@ mtev_lfu_create(int32_t max_entries, void (*free_fn)(void *))
   } else {
     hs_init(&r->hash, CK_HS_MODE_OBJECT | CK_HS_MODE_DELETE, lfu_entry_hash, hs_string_compare, max_entries * 2);
   }
-  STAILQ_INIT(&r->lfu_frequency_list);
+  TAILQ_INIT(&r->lfu_frequency_list);
   r->lfu_cache_size = 0;
   r->expire_count = 0;
   r->max_entries = max_entries;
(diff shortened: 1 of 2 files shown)


src/utils/mtev_lfu.c
```diff
@@ -11,7 +11,7 @@
 
 struct lfu_cache_entry;
 
-STAILQ_HEAD(frequency_list, lfu_cache_entry);
+TAILQ_HEAD(frequency_list, lfu_cache_entry);
```
rileyberton: need bidirectional queue

src/utils/mtev_lfu.c
```diff
@@ -234,21 +231,27 @@ touch_lfu_cache_no_lock(mtev_lfu_t *c, struct lfu_entry *e)
...
-        next_bucket == STAILQ_FIRST(&c->lfu_frequency_list) ||
+        next_bucket == TAILQ_FIRST(&c->lfu_frequency_list) ||
         next_bucket->frequency != bucket_freq + 1) {
       need_new_bucket = mtev_true;
     }
     struct lfu_cache_entry *empty = remove_from_frequency_list_no_lock(c, e);
     if (need_new_bucket) {
       if (empty != NULL) {
+        /* reuse this bucket instead of allocating */
+        empty->frequency = bucket_freq + 1;
```
rileyberton: There may be gaps in the bucket list.

`1 -> 3 -> 7`

If we are in `3` and `3` becomes empty.. when it becomes 4 it needs to be inserted *after* `1`, not at the HEAD which would create the wrong order.

Now we check next and prev buckets.. if we have advanced past `next_bucket` then we treat it as the prior, if not, we treat `prev_bucket` as prior.  `prev_bucket` will always be lower in frequency by definition.

src/utils/mtev_lfu.c
```diff
@@ -234,21 +231,27 @@ touch_lfu_cache_no_lock(mtev_lfu_t *c, struct lfu_entry *e)
...
+        if (next_bucket && next_bucket->frequency < empty->frequency) {
+          insert_to_bucket_list(c, empty, next_bucket);
+        } else if (prev_bucket && prev_bucket->frequency < empty->frequency) {
```
rileyberton: Probably should just be `else` instead of `else if`...

src/utils/mtev_lfu.c
```diff
@@ -234,22 +231,30 @@ touch_lfu_cache_no_lock(mtev_lfu_t *c, struct lfu_entry *e)
...
+        next_bucket == TAILQ_FIRST(&c->lfu_frequency_list) ||
         next_bucket->frequency != bucket_freq + 1) {
       need_new_bucket = mtev_true;
     }
     struct lfu_cache_entry *empty = remove_from_frequency_list_no_lock(c, e);
     if (need_new_bucket) {
       if (empty != NULL) {
+        /* reuse this bucket instead of allocating */
+        empty->frequency = bucket_freq + 1;
+        if (next_bucket && next_bucket->frequency < empty->frequency) {
```
aidancully: this case should be impossible? `c->lfu_frequency_list` is now stored sorted, so `empty`'s old frequency is guaranteed to be less than `next_bucket->frequency`. Further, we've already checked if `next_bucket->frequency` == new frequency (i.e. old frequency + 1), and it's not, because we `need_new_bucket`... I think we're OK w/ replacing all this logic with the body of the `else` block.

rileyberton: yes, I suppose either next_bucket is equal to the new frequency or greater.. It can't be less. I will simplify further.

PR #303, opened 2017-08-16
More coverity fixes

SamuelWM01:
Fixed a few more issues in coverity. Need to push and merge this so I can do a git pull and Fix a few more remaining changes.
Reviews appreciated.
Edit to lines 85 & 86 of mtev_cpuid.c is an attempt to have at least one condition reach the default return and avoid a deadcode error.

### DIFF
--- a/src/modules/lua_general.c
+++ b/src/modules/lua_general.c
@@ -876,9 +876,9 @@ lua_repl_prompt(EditLine *el) {
   conf = get_config(info->self);
   lmc = pthread_getspecific(conf->key);
 
-  if(!lmc||!pthread_equal(eventer_get_owner(ncct->e), pthread_self()))
+  if(!lmc || !pthread_equal(eventer_get_owner(ncct->e), pthread_self()))
     snprintf(info->prompt, sizeof(info->prompt), "lua_general(...)# ");
-  else 
+  else
     snprintf(info->prompt, sizeof(info->prompt), tl, get_eventer_id(ncct), lmc->lua_state);
   return info->prompt;
 }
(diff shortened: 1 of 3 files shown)


src/modules/lua_general.c
```diff
@@ -878,8 +878,8 @@ lua_repl_prompt(EditLine *el) {
 
   if(!pthread_equal(eventer_get_owner(ncct->e), pthread_self()))
     snprintf(info->prompt, sizeof(info->prompt), "lua_general(...)# ");
-  else
-    snprintf(info->prompt, sizeof(info->prompt), tl, get_eventer_id(ncct), lmc->lua_state);
+  else 
+    if(lmc)  snprintf(info->prompt, sizeof(info->prompt), tl, get_eventer_id(ncct), lmc->lua_state);
```
postwait: This would allow us returning without filling out info->prompt.

This shold be moved up to the first if:

``` if(!lmc || !pthread_equal(..... ```

and the else remain as it was.

Comment (SamuelWM01) [inline src/modules/lua_general.c]: okay

File: src/modules/lua_general.c [hunk 2]
```diff
@@ -876,9 +876,9 @@ lua_repl_prompt(EditLine *el) {
   conf = get_config(info->self);
   lmc = pthread_getspecific(conf->key);
 
-  if(!pthread_equal(eventer_get_owner(ncct->e), pthread_self()))
+  if(!lmc||!pthread_equal(eventer_get_owner(ncct->e), pthread_self()))
     snprintf(info->prompt, sizeof(info->prompt), "lua_general(...)# ");
-  else
+  else
```
gallison93: added a trailing space at the end this line, should remove it

src/modules/lua_general.c
```diff
@@ -876,9 +876,9 @@ lua_repl_prompt(EditLine *el) {
   conf = get_config(info->self);
   lmc = pthread_getspecific(conf->key);
 
-  if(!pthread_equal(eventer_get_owner(ncct->e), pthread_self()))
+  if(!lmc||!pthread_equal(eventer_get_owner(ncct->e), pthread_self()))
```
gallison93: add spaces around the OR. `if(!lmc || !pthread...`

PR #304, opened 2017-08-16
Even More Coverity Fixes

SamuelWM01:
Fixes for an insecure temp file, deadcode, and unchecked return values.
Couldn't fix other temp file yet due to testing errors.
**Warning** umask might break stuff due to the nature of the command.  
Merge with caution and if a problem is noticed, notify me and I will fix it.


SamuelWM01:
tests worked and libmtev built with the temp file insecurity fix, I'm just concerned that the tests may not be adequate for the change 🤔

rileyberton:
Looks good to me

PR #306, opened 2017-08-17
Failure Fix

SamuelWM01:
Attempted fix to insecure temp file caused failures.
Changing it back for the time being.

### DIFF
--- a/src/utils/mtev_stacktrace.c
+++ b/src/utils/mtev_stacktrace.c
@@ -119,7 +119,9 @@ void mtev_stacktrace(mtev_log_stream_t ls) {
     /* This is Async-Signal-Safe (at least on Illumos) */
     char tmpfilename[MAXPATHLEN];
     snprintf(tmpfilename, sizeof(tmpfilename), "/var/tmp/mtev_%d_XXXXXX", (int)getpid());
+    int oldmask = umask(0600);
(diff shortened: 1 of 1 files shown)


esproul:
```
utils/mtev_stacktrace.c:122: error: 'oldmask' undeclared (first use in this function)
```

Failures only happened on Linux, and not on OmniOS, but according to Sam, simply declaring the `oldmask` variable beforehand did not work.

SamuelWM01:
When I had the variable declared before hand it crashed and burned in my devzone, looking into it rn.

src/utils/mtev_stacktrace.c
```diff
@@ -119,9 +119,7 @@ void mtev_stacktrace(mtev_log_stream_t ls) {
     /* This is Async-Signal-Safe (at least on Illumos) */
     char tmpfilename[MAXPATHLEN];
     snprintf(tmpfilename, sizeof(tmpfilename), "/var/tmp/mtev_%d_XXXXXX", (int)getpid());
-    oldmask = umask(0600);
```
postwait: Declare it here... 

```
int oldmask = umask(0600);
```

SamuelWM01: just did that, It seems to be working in my devzone so I'll push it back up.

PR #308, opened 2017-08-18
Mtev hash iterate spmc

rileyberton:
This relies on HEAD of concurrency kit to get `ck_hs_next_spmc`

I will leave this open until we have a libck package based on HEAD.

There is a bunch of safe memory stuff below.  

The short story is: You can use mtev_hash as SPMC going forward and iterate/get from it on the read side without locks.  Even in the face of deletes from the writer thread.  Before this merge this wasn't really safe because ck_hs did not memoize the internal map and iterations and gets could hit the old map after a resize in the write thread which could have already been freed.


rileyberton:
Yes, did this.

PR #310, opened 2017-08-22
Protect against null dereference.

postwait:
It is possible to release the head chain (nulling it), then break to here
from the EAGAIN read state to if there is data available and wind up with
no next chunk and head unassigned.  We need to always assign head upon
entering this section of code.


gallison93:
LGTM

PR #313, opened 2017-08-28
fix return value of mtev_hash_set to indicate whether something was r…

rileyberton:
…eplaced and utilize this in mtev_hash_replace for free function calling

### DIFF
--- a/src/utils/mtev_hash.h
+++ b/src/utils/mtev_hash.h
@@ -48,11 +48,9 @@ typedef enum mtev_hash_lock_mode {
   MTEV_HASH_LOCK_MODE_SPIN = 2
 } mtev_hash_lock_mode_t;
 
-typedef enum mtev_hash_return_code {
-  MTEV_HASH_FAILURE=0,
-  MTEV_HASH_SUCCESS=1,
-  MTEV_HASH_SUCCESS_REPLACEMENT=2
-} mtev_hash_return_code_t;
+#define MTEV_HASH_FAILURE 0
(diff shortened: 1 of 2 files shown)


src/utils/mtev_hash.h
```diff
@@ -142,6 +147,10 @@ int mtev_hash_replace(mtev_hash_table *h, const char *k, int klen, void *data,
                       NoitHashFreeFunc keyfree, NoitHashFreeFunc datafree);
 
 /* replace and return the old value and old key that was in this hash location
+ * 
+ * will return 1 on successful set with no replacement
+ * will return 0 on failure to set
+ * will return 2 on successful set with replacement
  */
 int mtev_hash_set(mtev_hash_table *h, const char *k, int klen, void *data,
```
aidancully: what do you think of changing the return type to `mtev_hash_return_code_t`? then you won't need to document that `1` means `set with no replacement`, etc..

rileyberton: I don't want to make this function incompatible with the rest of the file and I don't want to change everything which will introduce compilation problems in all the C++ code that uses a hash.

aidancully: It'd only affect the C++ code that calls `mtev_hash_set`, I haven't seen if we've added much of that, so this is still my preference, but your call. If you leave it as is, please change `will return 1` to `will return MTEV_HASH_SUCCESS` etc. And I guess in that case, I'd replace the enum with a set of `#define`s.

aidancully:
LGTM

PR #315, opened 2017-08-29
Add Missing Release In mtev_conf_security_init

pamaddox:
We were failing to release "secnode" here, causing a potential deadlock.


aidancully:
lgtm

PR #316, opened 2017-08-31
Fix Crashing Bug In lua_mtev.c

pamaddox:
When we free, we need to set the eptr to NULL to prevent reuse.


pamaddox:
Agreed with Aidan.

aidancully:
AFAICT, C code can only see `cl->eptr` through `cl`.`cl->eptr` was allocated by `lua_newuserdata`, so Lua also has access to `cl->eptr` (in a way), and that's what's causing the UAF. In the problem case I looked at, C code has free'd the `eventer_t` via `mtev_lua_resume_clean_events`, but `cl->eptr` was left populated. So when Lua attempted to GC `cl->eptr`, we got a UAF.

PR #320, opened 2017-09-18
add `eventer_at_next_opportunity` and `eventer_add_at_next_opportunit…

aidancully:
…y` functions.

these wake up the relevant event-loop, to call the requested callback
as soon as possible.

### DIFF
--- a/src/eventer/eventer.h
+++ b/src/eventer/eventer.h
@@ -906,6 +906,23 @@ eventer_at(eventer_func_t func, void *cl, struct timeval t) {
   return e;
 }
 
+/*! \fn eventer_t eventer_at_next_opportunity(eventer_func_t func, void *closure, pthread_t owner)
+    \brief Convenience function to create an event to run a callback on a specific thread.
+    \param func the callback function to run.
+    \param closure the closure to be passed to the callback.
+    \param owner the event-loop thread on which to run the callback.
+    \return an event that has not been added to the eventer.
+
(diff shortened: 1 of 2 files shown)


src/eventer/eventer.h
```diff
@@ -906,6 +906,23 @@ eventer_at(eventer_func_t func, void *cl, struct timeval t) {
...
+    \brief Convenience function to create an event to run a callback on a specific thread.
+    \param func the callback function to run.
+    \param closure the closure to be passed to the callback.
+    \param owner the event-loop thread on which to run the callback.
+    \return an event that has not been added to the eventer.
+
+    > Note this does not actually schedule the event. See [`eventer_add_at_next_opportunity`](c.md#eventeraddatnextopportunity).
+*/
+static inline eventer_t
+eventer_at_next_opportunity(eventer_func_t func, void *cl, pthread_t owner) {
```
rileyberton: maybe `eventer_alloc_at_next_opportunity`?  The other creation functions include `alloc` in the name.

aidancully: the naming as based on `eventer_at` and `eventer_add_at`, which behave very similarly to these functions..

rileyberton: tis fine, just differs from the other alloc functions.. I would argue that `eventer_at` should be called `eventer_alloc_at`.. but we can fix in the future.

PR #324, opened 2017-09-26
Add Function To Run Commands Synchronously And Return Output

pamaddox:
In mtevbusted.... we need a way to run an external process and return
the output of said process. Added functions to allow for this.

### DIFF
--- a/test/mtevbusted/child.lua
+++ b/test/mtevbusted/child.lua
@@ -18,7 +18,6 @@ function mk_coroutine_reader(start, done, fd)
       end
       output = output .. line
     end
-    print("END")
     mtev.notify(done, output)
     fd:close()
     return nil
--- a/test/mtevbusted/child.lua
+++ b/test/mtevbusted/child.lua
@@ -96,8 +96,16 @@ function TestProc:capturecommand(props)
   local stdout_key, stdout_data = mtev.waitfor(self.stdout, self.timeout)
   local stderr_key, stderr_data = mtev.waitfor(self.stderr, self.timeout)
 
-  self.proc:kill()
-  self.proc:wait(10)
+  local ret = self.proc:wait(1)
+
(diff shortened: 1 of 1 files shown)


test/mtevbusted/child.lua
```diff
@@ -57,6 +57,42 @@ function TestProc:waitfor(key, timeout)
...
+  end
+  mtev.notify(self.start, true)
+  local key, data = mtev.waitfor(self.output,self.timeout)
+
+  self.proc:kill()
+  self.proc:wait(10)
+  self.proc = nil
+
+  return data
+end
```
aidancully: I think this should capture the exit status (from `self.proc:wait()`), the standard output (like you have) and the stderr output. Also, `self.proc:wait(10)` - I think the `10` should be a parameter, maybe in `props`?

pamaddox: self.proc:wait(10) is a direct copy/paste from the "start()" function - if we change it here, we'd probably want to change it there as well so we don't have inconsistent behavior.

As for capturing stderr - would have to look into that a bit, not sure how that'd work with the current structure to read data/notice errors.

pamaddox: Should be addressed now

test/mtevbusted/child.lua
```diff
@@ -57,6 +76,41 @@ function TestProc:waitfor(key, timeout)
...
+    self.proc:kill()
+    ret = self.proc:wait(10)
+  end
+  if ret == nil then
+    error("Couldn't quit process")
+  end
+
+  self.proc = nil
+
+  return stdout_data, stderr_data
```
aidancully: `return ret, stdout_data, stderr_data`

aidancully:
LGTM, but this is more @postwait's area

PR #327, opened 2017-10-11
Add sha256 functions to mtev_lua

esproul:


esproul:
@postwait just to clarify, since I don't see any of the existing hash functions documented, this will be in the "Lua" section of the API reference, i.e. http://circonus-labs.github.io/libmtev/apireference/#lua ?

postwait:
There has been a document as you touch things approach.  See other documented lua functions in the code.

PR #328, opened 2017-10-18
using umask on stacktrace is not safe.

aidancully:
other threads may rely on umask having an application-set value,
so that modifying it during a stack-dump can cause those other threads
to misbehave.


pamaddox:
LGTM

PR #333, opened 2017-10-24
mtev_sem_wait_noeintr function

aidancully:


aidancully:
anyone? if this goes too much against the grain, i'd prefer it be closed than to hang out open and pending.

PR #335, opened 2017-10-27
support non-string keys in mtev_lfu

rileyberton:
store key_len with the LFU key so we can store arbitrary non-string keys.

### DIFF
--- a/src/utils/mtev_lfu.c
+++ b/src/utils/mtev_lfu.c
@@ -64,7 +64,7 @@ static struct ck_malloc malloc_ck_hs = {
 static unsigned long
 lfu_entry_hash(const void *k, unsigned long seed)
 {
-  struct lfu_key *key = (struct lfu_key *)k;
+  const struct lfu_key *key = (const struct lfu_key *)k;
   return mtev_hash__hash(key->key, key->key_len, seed);
 }
 
@@ -85,8 +85,8 @@ hs_init(ck_hs_t *hs, unsigned int mode, ck_hs_hash_cb_t *hf, ck_hs_compare_cb_t 
 static bool
 hs_lfu_key_compare(const void *a, const void *b)
 {
-  struct lfu_key *left = (struct lfu_key *)a;
-  struct lfu_key *right = (struct lfu_key *)b;
+  const struct lfu_key *left = (const struct lfu_key *)a;
+  const struct lfu_key *right = (const struct lfu_key *)b;
 
   if (left->key_len < right->key_len) return -1;
   if (left->key_len > right->key_len) return 1;
--- a/src/utils/mtev_lfu.c
+++ b/src/utils/mtev_lfu.c
@@ -8,6 +8,7 @@
 #include <stddef.h>
 
 #define GC_CADENCE 10000
+#define ALLOCA_LIMIT 1024
 
 struct lfu_cache_entry;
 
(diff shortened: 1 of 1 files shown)


src/utils/mtev_lfu.c
```diff
@@ -78,19 +83,25 @@ hs_init(ck_hs_t *hs, unsigned int mode, ck_hs_hash_cb_t *hf, ck_hs_compare_cb_t
 }
 
 static bool
-hs_string_compare(const void *a, const void *b)
+hs_lfu_key_compare(const void *a, const void *b)
 {
-  return strcmp((const char * const)a, (const char * const)b) == 0;
+  struct lfu_key *left = (struct lfu_key *)a;
+  struct lfu_key *right = (struct lfu_key *)b;
```
aidancully: nit (and I know it's a little silly), but would prefer if the vars and the typecasts all said `const`.

src/utils/mtev_lfu.c
```diff
@@ -305,8 +319,12 @@ mtev_lfu_get(mtev_lfu_t *c, const char *key, size_t key_len)
     return NULL;
   }
 
-  unsigned long hash = CK_HS_HASH(&c->hash, lfu_entry_hash, key);
-  void *entry = ck_hs_get(&c->hash, hash, key);
+  struct lfu_key *tempkey = alloca(sizeof(struct lfu_key) + key_len);
```
pamaddox: If key_len is not bounded, may want to put a size limit on it, lest you alloca() something infinitely huge.... like, for example:

if (key_len < 65535) /* use alloca */
else /* use malloc/free */

src/utils/mtev_lfu.c
```diff
@@ -78,19 +83,25 @@ hs_init(ck_hs_t *hs, unsigned int mode, ck_hs_hash_cb_t *hf, ck_hs_compare_cb_t
 }
 
 static bool
-hs_string_compare(const void *a, const void *b)
+hs_lfu_key_compare(const void *a, const void *b)
```
pamaddox: not sure if this is right.... it would return `b` < `aa`, and I'm not sure that's what we want.

src/utils/mtev_lfu.c
```diff
@@ -326,11 +344,14 @@ mtev_lfu_remove(mtev_lfu_t *c, const char *key, size_t key_len)
   if (c->max_entries == 0) {
     return NULL;
   }
+  struct lfu_key *tempkey = alloca(sizeof(struct lfu_key) + key_len);
```
pamaddox: Same as above

src/utils/mtev_lfu.c
```diff
@@ -269,16 +280,19 @@ mtev_lfu_put(mtev_lfu_t *lfu, const char *key, size_t key_len, void *val)
     return mtev_false;
   }
 
-  unsigned long hash = CK_HS_HASH(&lfu->hash, lfu_entry_hash, key);
+  struct lfu_key *tempkey = alloca(sizeof(struct lfu_key) + key_len);
+  tempkey->key_len = key_len;
+  memcpy(tempkey->key, key, key_len);
```
aidancully: note: you could avoid the `alloca` and `memcpy` if you changed the key type to include a pointer, as well as the flexible array element at the end... Lookups / compares would use the pointer field, and when stored in the heap, the pointer would point to the flexible array element.

PR #336, opened 2017-11-07
move LRU to non-string keys, introduce memory safety through a get/re…

rileyberton:
…lease semantic

### DIFF
--- a/src/utils/mtev_lfu.c
+++ b/src/utils/mtev_lfu.c
@@ -21,6 +21,7 @@ struct lfu_key {
 
 struct lfu_entry {
   void *entry;
+  uint64_t ref_cnt;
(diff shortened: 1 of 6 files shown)


src/utils/mtev_lfu.c
```diff
@@ -350,12 +349,32 @@ mtev_lfu_get(mtev_lfu_t *c, const char *key, size_t key_len)
     struct lfu_entry *r = container_of(entry, struct lfu_entry, key);
     touch_lfu_cache_no_lock(c, r);
     pthread_mutex_unlock(&c->mutex);
```
postwait: Can another thread perform an eviction on this line of code making the subsequent inc_64 "too late"?

PR #339, opened 2017-11-09
Allow intermediate nodes in eventer/config

HeinrichHartmann:
So that we will find eventer/config stanzas in include files.


HeinrichHartmann:
obsolete.

PR #342, opened 2017-11-13
Support mtev.stats to register and record stats.

postwait:


src/modules/lua_mtev_stats.c
```diff
@@ -0,0 +1,270 @@
...
+  handle = *udata;
+  type = stats_handle_type(handle);
+  if(lua_isnil(L,2)) {
+    /* This clears a value */
+    stats_set(handle, type, NULL);
+    return 0;
+  }
+  switch(type) {
+    case STATS_TYPE_STRING:
+      stats_set(handle, STATS_TYPE_STRING, (char *)lua_tostring(L,2));
```
aidancully: recommend verifying that argument is a string before calling `stats_set`. similar comment applies to all numeric types.

postwait: This was actually intentional as it seems more lua-esque to auto convert when possible like other functions do.  It allos setting strings like 1234 and having them be "1234" and someone setting a "9.2" number and having it flow through as 9.2.

PR #345, opened 2017-11-15
Lua add _unwarp_json

HeinrichHartmann:


src/modules/lua_mtev.c
```diff
@@ -3405,6 +3405,22 @@ nl_tojson(lua_State *L) {
...
+// Removes wrapping around mtev_json_object structure, and leaves a
+// mtev_json_object* on the lua stack, so other C functions can make
+// use of it.
+static int
+nl__unwrap_json(lua_State *L){
+  json_crutch **docptr;
+  mtev_json_object **unwrapped;
+  if (lua_gettop(L) != 1) luaL_error(L, "_unwrap_json requires one argument");
+  docptr = (json_crutch **) lua_touserdata(L, 1);
+  unwrapped = (mtev_json_object **) lua_newuserdata(L, sizeof(mtev_json_object*));
```
HeinrichHartmann: it seems to me that `lightuserdata` would be a more natural choice for these structures (json_crutch/mtev_json_object).

But apparently, luajit on OmniOS does not support lighuserdata (for stack allocations at least):
http://lua.2524044.n2.nabble.com/LPEG-on-Solaris-td7672860.html

Is there another reason, we are not using lightuserdata I am missing here?

aidancully: Could be wrong, but AIUI, lightuserdata doesn't get GC'ed as nicely as userdata does? and mtev.json objects need to be specially GC'ed, since these objects are reference-counted in C.

postwait: You don't control when the docptr userdata will be GCd.  So this isn't safe.  You unwrap the json object into a new userdata, but don't increment it's refcnt.  To do this we'd need to have a jsonobject metatable with a __gc function the "puts" the object and then we'd "get" here in the assignment to *unwrapped.

HeinrichHartmann:
That's the reason there are two commits in this PR. Just click on them to see the individual diffs.

HeinrichHartmann:
LGTM.

Did some local testing. This appears to be working as expected.

PR #346, opened 2017-11-16
Track errno for failed asynch writes

postwait:


src/utils/mtev_log.c
```diff
@@ -572,9 +573,12 @@ posix_logio_asynch_write(asynch_log_ctx *actx, asynch_log_line *line) {
   struct posix_op_ctx *po;
   int rv = -1;
   po = actx->userdata;
-  if(po && po->fd >= 0)
+  actx->last_errno = 0;
+  if(po && po->fd >= 0) {
     rv = write(po->fd, line->buf_dynamic ? line->buf_dynamic : line->buf_static,
                line->len);
+    if(rv < 0) actx->last_errno = errno;
```
aidancully: also record `jlog_ctx_errno` from `jlog_logio_asynch_write`?

aidancully: also applies to `asynch_thread_create`... still looking for other cases.

aidancully: `posix_logio_write` can call `write` directly, probably want to set there, too. `posix_logio_close` for completeness... don't see other possible places to set the value.

PR #348, opened 2017-11-28
Fix Memory Leak In nl_waitfor Functions

pamaddox:
We were basically leaking eventer objects all over the place due to not
cleaning them up when we resume. Fix this memory leak.

### DIFF
--- a/src/modules/lua_mtev.c
+++ b/src/modules/lua_mtev.c
@@ -1565,6 +1565,8 @@ nl_waitfor_notify(lua_State *L) {
     mtevAssert(ci);
     mtevAssert(eventer_remove(q->pending_event));
     mtev_lua_deregister_event(ci, q->pending_event, 0);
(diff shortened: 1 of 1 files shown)


src/modules/lua_mtev.c
```diff
@@ -1619,7 +1624,7 @@ nl_waitfor(lua_State *L) {
     mtev_hash_store(ci->lmc->pending, q->key, strlen(q->key), q);
   } else {
     q = vptr;
-    if(q->L) luaL_error(L, "waitfor cannot be called concurrently");
+    if(q->L || q->pending_event) luaL_error(L, "waitfor cannot be called concurrently");
```
postwait: q->L is sufficient I believe to prevent coroutine.

pamaddox: Fixed

PR #354, opened 2017-12-07
V1

postwait:


esproul:
What is the purpose of the "improved tag detection"?  `git tag` returns the full list of tags, so I don't understand how loading that into `SYM` makes sense.

PR #356, opened 2017-12-11
Fix Some Memory Leaks And Races

pamaddox:
We could potentially leak memory if mtev_hash_store calls fail, and a
cleanup function was not freeing everything it should have been. Fixed
these conditions.


pamaddox:
lmc is thread-safe, so we don't need all this checking. PR is unnecessary. Withdrawing.

PR #357, opened 2017-12-13
Fix Leaks And Add Safety To Lua Functions

pamaddox:
* Fix a memory leak where we weren't freeing the data in a hash
* Check for failure on mtev_hash_store and if we DO fail, clean up the
  allocated memory and throw an error.


rileyberton:
Looks good to me

PR #363, opened 2018-01-25
Make SONAME be just the major version

esproul:
Make it simpler to introduce backward-compatible new features, which in semver is the minor release.  We plan to reserve ABI-breaking changes for major releases.

### DIFF
--- a/configure.in
+++ b/configure.in
@@ -98,7 +98,7 @@ CPPFLAGS="$CPPFLAGS -D_REENTRANT "'-I$(top_srcdir)/src'
 DTRACEHDR=libmtev_dtrace_probes.h
 DTRACEHDR_TRANSFORM="copy"
 DOTSO=.so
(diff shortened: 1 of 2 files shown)


esproul:
Tested on OmniOS:

```
$ ls -l /tmp/mtevtest/opt/circonus/lib/amd64/
total 6896
lrwxrwxrwx   1 root     root          16 Jan 25 19:58 libmtev.so -> libmtev.so.1.0.2
lrwxrwxrwx   1 root     root          16 Jan 25 19:58 libmtev.so.1 -> libmtev.so.1.0.2
-rwxr-xr-x   1 root     root     3502200 Jan 25 19:58 libmtev.so.1.0.2

$ elfdump -d /tmp/mtevtest/opt/circonus/lib/amd64/libmtev.so.1.0.2  | grep SONAME
      [29]  SONAME            0xc564              libmtev.so.1
```

PR #365, opened 2018-01-29
Use 64bit integer conf_get to set max_size and retain_size log options.

postwait:


rileyberton:
Looks good to me

esproul:
FYI, this came up because of a configuration where `retain_bytes` was set to > `INT32_MAX`, causing the value to be negative, and therefore retention was not enforced and a filesystem filled up.

PR #367, opened 2018-02-12
Add try variant of eventer_add_asynch APIs

postwait:


src/eventer/eventer_jobq.c
```diff
@@ -414,6 +405,44 @@ eventer_jobq_enqueue(eventer_jobq_t *jobq, eventer_job_t *job, eventer_job_t *pa
...
+     uint32_t bl;
+     do {
+       bl = ck_pr_load_32(&jobq->backlog);
+       if(bl >= jobq->max_backlog) {
+         free(job);
+         return mtev_false;
+       }
+     } while(!ck_pr_cas_32(&jobq->backlog, bl, bl+1));
+   }
+   else ck_pr_inc_32(&jobq->backlog);
```
rileyberton: I don't see where the backlog is decremented.

postwait: When the job "goes live"

```
  if(job) {
    job->next = NULL; /* To reduce any confusion */
    if(job->fd_event) ck_pr_dec_32(&jobq->backlog);
    ck_pr_inc_32(&jobq->inflight);
  }
```

postwait: (this isn't new behavior).

In the original enqueue we upped total_jobs and backlog.

Now we split the increment so in the `internal` version only total_jobs is incremented and in the calling wrapper we're more careful w.r.t. incrementing backlog.

PR #371, opened 2018-02-20
Fix non-Linux builds after PR#370

esproul:
Not sure if this is the best/right way to solve it but it restores a working build and passes tests on OmniOS.

### DIFF
--- a/src/mtev_defines.h
+++ b/src/mtev_defines.h
@@ -216,8 +216,9 @@ static inline void uuid_unparse_lower(uuid_t in, char *out) {
 #define portable_readdir_r(a,b,c) (((*(c)) = readdir_r(a,b)) == NULL)
 #else
 /* https://lwn.net/Articles/696474/ */
-#if defined(linux) || defined(__linux) || defined(__linux__)
-#if defined(__GLIBC_PREREQ) && __GLIBC_PREREQ(2, 24)
+/* https://lists.nongnu.org/archive/html/libunwind-devel/2011-11/msg00046.html */
+#if defined(__GLIBC__) && defined(__GLIBC_PREREQ)
+#if __GLIBC_PREREQ(2, 24)
 #if HAVE_DIRENT_H
 #include <dirent.h>
 #endif
--- a/src/mtev_defines.h
+++ b/src/mtev_defines.h
@@ -216,6 +216,7 @@ static inline void uuid_unparse_lower(uuid_t in, char *out) {
 #define portable_readdir_r(a,b,c) (((*(c)) = readdir_r(a,b)) == NULL)
 #else
 /* https://lwn.net/Articles/696474/ */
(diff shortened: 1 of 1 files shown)


esproul:
@postwait I tried combining but the preprocessor still complained:

```
diff --git a/src/mtev_defines.h b/src/mtev_defines.h
index 9a766dc..19fb0a4 100644
--- a/src/mtev_defines.h
+++ b/src/mtev_defines.h
@@ -216,7 +216,7 @@ static inline void uuid_unparse_lower(uuid_t in, char *out) {
 #define portable_readdir_r(a,b,c) (((*(c)) = readdir_r(a,b)) == NULL)
 #else
 /* https://lwn.net/Articles/696474/ */
-#if defined(__GLIBC_PREREQ) && __GLIBC_PREREQ(2, 24)
+#if (defined(linux) || defined(__linux) || defined(__linux__)) && defined(__GLIBC_PREREQ) && __GLIBC_PREREQ(2, 24)
 #if HAVE_DIRENT_H
 #include <dirent.h>
 #endif
```

Still got:

```
In file included from mtev_main.c:33:0:
mtev_defines.h:219:108: error: missing binary operator before token "("
 #if (defined(linux) || defined(__linux) || defined(__linux__)) && defined(__GLIBC_PREREQ) && __GLIBC_PREREQ(2, 24)
                                                                                                            ^
```

esproul:
Tested a better solution, suggested by @postwait, which was to split the define test for `__GLIBC_PREREQ` from the actual macro expansion.  Additionally we guard with a define condition for `__GLIBC__`.  This compiles and passes tests on both OmniOS r151014 and Ubuntu 17.10.

esproul:
This still isn't right, it breaks on glibc < 2.24 now.

```
- compiling mtev_rest.c
cc1: warnings being treated as errors
mtev_conf.c: In function 'mtev_conf_read_into_node':
mtev_conf.c:1102: error: implicit declaration of function 'portable_readdir_r'
```

src/mtev_defines.h
```diff
@@ -226,6 +228,7 @@ static inline int portable_readdir_r(DIR *dirp, struct dirent *entry, struct dir
   return readdir_r(dirp, entry, result);
 #pragma GCC diagnostic pop
 }
```
postwait:
```
#else 
#define portable_readdir_r readdir_r
```

esproul: I see, so we duplicate the `portable_readdir_r`, but seems unavoidable.

esproul:
Tests now pass on glibc < 2.24 (EL6)

cody271:
Sorry, I should have tested this more thoroughly!

postwait:
CI is king.  At some point, we'll do auto-CI on PRs like we do for some of our other projects.  Then we can all be more thorough without investing any more human cycles.

PR #373, opened 2018-02-20
Minor fixes and cleanups to mtev_cluster.c

HeinrichHartmann:
See commit messages.

### DIFF
--- a/src/mtev_cluster.c
+++ b/src/mtev_cluster.c
@@ -1096,12 +1096,14 @@ mtev_cluster_to_xmlnode(mtev_cluster_t *c) {
   snprintf(maturity, sizeof(maturity), "%d", c->maturity);
   xmlSetProp(cluster, (xmlChar *)"maturity", (xmlChar *)maturity);
 
-  xmlNodePtr node;
-  char uuid_str[UUID_STR_LEN+1];
-  uuid_unparse_lower(c->oldest_node->id, uuid_str);
-  node = xmlNewNode(NULL, (xmlChar *)"oldest_node");
-  xmlSetProp(node, (xmlChar *)"uuid", (xmlChar *)uuid_str);
-  xmlAddChild(cluster, node);
+  if(c->oldest_node) {
+    xmlNodePtr node;
+    char uuid_str[UUID_STR_LEN+1];
+    uuid_unparse_lower(c->oldest_node->id, uuid_str);
+    node = xmlNewNode(NULL, (xmlChar *)"oldest_node");
+    xmlSetProp(node, (xmlChar *)"uuid", (xmlChar *)uuid_str);
+    xmlAddChild(cluster, node);
+  }
 
   for(i=0;i<c->node_cnt;i++) {
     mtev_cluster_node_t *n = &c->nodes[i];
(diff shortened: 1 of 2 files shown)


HeinrichHartmann:
Removed aesthetic changes.
Moved non-cluster stuff to #374.
Also added the drop_e fix to this PR.

src/mtev_cluster.c
```diff
@@ -801,19 +806,22 @@ mtev_cluster_set_self(uuid_t id) {
        my_id_str[UUID_STR_LEN+1];
   uuid_copy(my_cluster_id, id);
   uuid_unparse_lower(my_cluster_id, my_id_str);
-  if(!mtev_conf_get_stringbuf(MTEV_CONF_ROOT, "//clusters/@my_id",
-                              old_uuid, sizeof(old_uuid)) ||
-     strcmp(old_uuid, my_id_str)) {
+  int found_id = mtev_conf_get_stringbuf(MTEV_CONF_ROOT, "//clusters/@my_id",
+                                         old_uuid, sizeof(old_uuid));
+  if(!found_id || strcmp(old_uuid, my_id_str) != 0) {
```
postwait: I don't understand this change?  It seems to not change behavior, but make it look less like predicates elsewhere in libmtev.

src/mtev_cluster.c
```diff
@@ -1225,29 +1237,46 @@ mtev_cluster_init(void) {
...
+
+  // Set global cluster ID
+  if(mtev_conf_get_stringbuf(parent, "@my_id", my_id_str, sizeof(my_id_str))) {
+    int rv = uuid_parse(my_id_str, my_id);
+    if (rv != 0) {
+      mtevL(cerror, "Error parsing //clusters/@my_id: %s\n", my_id_str);
+    }
+    else {
+      mtev_cluster_set_self(my_id);
+      mtevL(cdebug,"//clusters/@myid was set to %s\n", my_id_str);
```
postwait: my_id

src/mtev_cluster.c
```diff
@@ -1225,29 +1237,46 @@ mtev_cluster_init(void) {
...
+    }
   }
+  else {
+    mtevL(cdebug,"//clusters/@myid not set.\n");
```
postwait: my_id

src/mtev_cluster.c
```diff
@@ -801,19 +806,22 @@ mtev_cluster_set_self(uuid_t id) {
...
     c = mtev_conf_get_section(MTEV_CONF_ROOT, "//clusters");
-    if(!mtev_conf_section_is_empty(c)) {
-      xmlNodePtr cnode = mtev_conf_section_to_xmlnodeptr(c);
-      xmlUnsetProp(cnode, (xmlChar *)"my_id");
-      xmlSetProp(cnode, (xmlChar *)"my_id", (xmlChar *)my_id_str);
-      CONF_DIRTY(c);
-      mtev_conf_mark_changed();
-      mtev_conf_request_write();
+    if (mtev_conf_section_is_empty(c)) {
+      mtevL(mtev_error, "Attempt to set_self with empty cluster secion\n");
```
postwait: s/secion/section/

src/mtev_cluster.c
HeinrichHartmann: This is related to a discussion in #syseng a few days ago: https://circonus.slack.com/archives/C0J5RAKDJ/p1518795659000234
I found this critical if statement very intransparent:
1. Side effects inside an if condition are confusing. First do something then check for errors.
2. The return codes of functions like mtev_conf_get_stringbuf, and strcmp, etc. are inconsistent. Does a non-zero value indicate success or error? No way to find this out without checking the docs/implementation, just "knowing" (which I do not). Here I named the results and makes the check more explicit.
For me the committed version is much more readable.

postwait: Consistency is king. It is standard practice throughout mtev code to use `strcmp`'s output as a boolean indicating nonmatch.  Also, almost everywhere in the code we use the boolean output of an mtev_conf_get function as success criteria.

This change deforms the code from its normal. I'm not supportive of this change.  It doesn't fix a bug and moves the code away from what is normal throughout the code base.

HeinrichHartmann: OK. Reverting.

src/mtev_cluster.c
```diff
@@ -1225,29 +1232,47 @@ mtev_cluster_init(void) {
...
+    if (rv != 0) {
+      mtevL(cerror, "Error parsing //clusters/@my_id: %s\n", my_id_str);
+    }
+    else {
+      mtev_cluster_set_self(my_id);
+      mtevL(cdebug,"//clusters/@myid was set to %s\n", my_id_str);
+    }
+  }
+  else {
+    mtevL(cdebug,"//clusters/@myid not set.\n");
```
postwait: my_id

PR #377, opened 2018-02-22
mtev_cluster: Make sure uuids are not null before printing

HeinrichHartmann:
# mtev_cluster: Make sure uuids are not null before printing.

It's not garuantted that my_cluster_id will be set at startup, so it can be nil.
Similarlly oldest_node is only determined after some heartbeats arrived.

# Complain about missing //clusters in mtev_cluster_set_self()

mtev_cluster_set_self fails silently if there are no matches for `//clusters`.
Now, it emit's a corresponding log line, and returns -1. 
The RC is not actually used on the call sites, but I found it clearer to indicate the error exit.
Happy to revert that to maintain the original signature (void return).

Also the logic was re-arranged a little:
1. Check for //clusters
2. Check for //clusters/@my_id
Not the other way around.

### DIFF
--- a/src/mtev_cluster.c
+++ b/src/mtev_cluster.c
@@ -798,27 +798,31 @@ mtev_cluster_size(mtev_cluster_t *c) {
   return c ? c->node_cnt : 0;
 }
 
-void
+int
 mtev_cluster_set_self(uuid_t id) {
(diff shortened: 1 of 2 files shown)


HeinrichHartmann:
Testing of clustering is not finished. This PR might be augmented.

PR #378, opened 2018-02-26
mtev_cluster PR-3

HeinrichHartmann:
See commit messages

### DIFF
--- a/src/mtev_cluster.c
+++ b/src/mtev_cluster.c
@@ -255,7 +255,7 @@ mtev_cluster_on_node_changed(mtev_cluster_t *cluster,
     if (compare_timeval(*new_boot_time, my_boot_time) == 0) {
       char node_name[128];
       mtev_cluster_node_to_string(sender, node_name, sizeof(node_name));
-      mtevL(mtev_debug,
+      mtevL(cdebug,
           "The following node in the cluster %s had the same startup time we have: '%s'\n",
           cluster->name, node_name);
       my_boot_time.tv_usec = random() % 1000000;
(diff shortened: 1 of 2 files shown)


src/mtev_capabilities_listener.c
```diff
@@ -83,13 +83,11 @@ mtev_capabilities_listener_init(void) {
...
 
-void
+int
 mtev_capabilities_add_feature(const char *feature, const char *version) {
   feature = strdup(feature);
   if(version) version = strdup(version);
-  if(!mtev_hash_store(&features, feature, strlen(feature), (void *)version))
-    mtevL(mtev_error, "Feature conflict! %s version %s\n",
-          feature, version ? version : "unpecified");
+  return mtev_hash_set(&features, feature, strlen(feature), (void *)version, NULL, NULL);
```
postwait: This is supposed to fail.

If you want to replace a feature, we need to add a new API.  mtev_capabilities_replace_feature.

Or... it doesn't fit as a feature and we should just query the cluster API.

HeinrichHartmann: yeah. I could not make this work without changes to mtev_capabilities_listener.
Changing `mtev_capabilities_add_feature` seemed less intrusive than creating an entirely new function.

The choice is up to you. I can add a "mtev_capabilities_add_or_replace_feature" function or leave it out entirely. No strong opinion here.

postwait:
Let's leave it out for now

HeinrichHartmann:
OK. _removed last commit_.

PR #379, opened 2018-03-11
Prevent javascript from adding thousands of lines of log to Internals…

vasuraman11:


vasuraman11:
I think the libmtev build is creating the minimized version of the js'es but I also saw it seems the minimized file is also committed.  Should I try to run the build and commit the minimized version, or should I let this be done by CI?

postwait:
CI won't do that.

rileyberton:
Yeah, I think we need to compile and check in the minimized version

vasuraman11:
This PR relates to part of CIRC-1525 (addressing the responsiveness issues but not collapse/expand).

vasuraman11:
Merging the fix and closing.

PR #385, opened 2018-04-04
Implement mtev.log_enabled(facility) in lua.

postwait:


HeinrichHartmann:
LGTM.

postwait:
Okay... technically if you supply an invalid log it will dynamically create it within libmtev and disable it.  So it wouldn't error.  I'll coalesce on the nil return always thing.

cody271:
The `lua_gettop` call is no longer necessary, but LGTM now!

PR #387, opened 2018-04-10
Updates to mtev_cluster, mtev_lua, mtev_conf

HeinrichHartmann:
See commit messages for more details.

### DIFF
--- a/src/mtev_cluster.c
+++ b/src/mtev_cluster.c
@@ -1304,9 +1304,9 @@ mtev_cluster_get_oldest_node(const mtev_cluster_t *cluster) {
 }
 
 mtev_boolean
-mtev_cluster_am_i_oldest_node(const mtev_cluster_t *cluster) {
+mtev_cluster_am_i_oldest_visible_node(const mtev_cluster_t *cluster) {
   if (cluster == NULL || cluster->oldest_node == NULL)
-    return mtev_false;
+    return mtev_true;
   return uuid_compare(cluster->oldest_node->id, my_cluster_id) == 0;
 }
 
--- a/src/modules/lua_mtev.c
+++ b/src/modules/lua_mtev.c
@@ -4014,113 +4014,6 @@ nl_watchdog_timeout(lua_State *L) {
   return 1;
 }
 
-static void
-mtev_lua_push_cluster_node(lua_State *L, mtev_cluster_node_t *node) {
-  char uuid_str[UUID_PRINTABLE_STRING_LENGTH];
-  if(node == NULL) {
-    lua_pushnil(L);
-  } else {
-    uuid_t nodeid;
-    mtev_cluster_node_get_id(node, nodeid);
-    uuid_unparse_lower(nodeid, uuid_str);
-
-    lua_createtable(L, 0, 3);
-
-    lua_pushstring(L, "id");
-    lua_pushstring(L, uuid_str);
-    lua_settable(L, -3);
-
-    lua_pushstring(L, "boot_time");
-    mtev_lua_push_timeval(L, mtev_cluster_node_get_boot_time(node));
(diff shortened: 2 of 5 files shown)


src/mtev_cluster.c
```diff
@@ -131,6 +131,9 @@ struct mtev_cluster_t {
   int64_t config_seq;
   int node_cnt;
   mtev_cluster_node_t *nodes;
+  // [2018-04-10 HH] cht is a concurrent hash-table mapping (char*) => (void*).
+  // It is initialized to hold a map node-UUID-string => mtev_cluster_node_t of known cluster
+  // nodes, and is used by the methods mtev_cluster_do_i_own, mtev_cluster_filter_owners.
```
postwait: It's not a concurrent hash table, it's a consistent hash table.  It also doesn't hold a map as you describe.  I'd simply remove this comment block.

src/mtev_cluster.c
```diff
@@ -627,7 +630,7 @@ int mtev_cluster_update_internal(mtev_conf_section_t cluster) {
      strlcpy(bufstr, "200", sizeof(bufstr));
   }
   period = strtoll(bufstr, &endptr, 10);
-  if(*endptr || period < 0 || period > 5000) {
+  if(*endptr || period < 0 || period > 30000) {
```
postwait: That's a really high period for a cluster. 30s?  I think 5s is a reasonably high bound for period.  What's the motivation here?

HeinrichHartmann: I upped this for debugging purposes. Logs quickly get spammed if you have a lot of debugging statements firing at every 5sec. Clearly you don't want to run like this in prod.

Also, I don't see harm in allowing such periods.

src/mtev_cluster.c
postwait: The cluster is represented as a consistent hash ring (like snowth).  When I have a 30 node cluster and 300k checks, which nodes runs which checks?  For pull checks only one can run them, so they need to collectively decide which nodes own which check data given the current availability.  This is what the CHT accomplishes.

src/mtev_cluster.c
postwait: Clusters with such high periods tend to behave in ways very surprising to the operator.   The dial on the instrument shouldn't go far beyond useful.  Our period's in prod should likely be around 500ms.

HeinrichHartmann: OK. Reverting.

src/mtev_cluster.c
HeinrichHartmann: OK. I understand the intentions here, but:
a) I don't see where a mapping `node-uuid => check-run-on-that-node` is implemented and maintained. Is that TBD? or am I missing something here. The only place I can see where cht is written to is in the `mtev_cluster_compile` function linked above.
b) Am I reading this code wrong, or are we really storing a table "node-uuid-str => node struct" in the cht, and not a map "node-uuid-str => check-run-on-that-node"
```
  *(cluster->cht) = mtev_cht_alloc();
  nodes = calloc(sizeof(*nodes), cluster->node_cnt);
  for(i=0; i<cluster->node_cnt; i++) {
    char uuid_str[UUID_STR_LEN+1];
    uuid_unparse_lower(cluster->nodes[i].id, uuid_str);
    nodes[i].name = strdup(uuid_str);
    nodes[i].userdata = &cluster->nodes[i];
  }
mtev_cht_set_nodes(*(cluster->cht), cluster->node_cnt, nodes);
```

HeinrichHartmann:
Removed the controversial bit's as requested. Merging the PR now.

I would still appreciate an answer to this Q:
https://github.com/circonus-labs/libmtev/pull/387#discussion_r180757671

PR #389, opened 2018-04-19
Support `mtev.timezone`

postwait:


HeinrichHartmann:
Pushed a nit to the branch. lua binding looks good otherwise.

HeinrichHartmann:
The tz information is read from the `/usr/share/zoneinfo`. Is this file guaranteed to be there? How is it kept up to date (leapseconds, etc. change)? Do the supported TZ zones differ from platform to platform?

Maybe it's worth to add a check function, that tests for the presence of the file, and a set of timezones that we can call on startup.

https://github.com/circonus-labs/libmtev/pull/389/files#diff-9ddf6c97c258d08cbe3203e87cd44cedR63

HeinrichHartmann:
Is there a reason the function returns values are sometimes in a separate line and sometimes not:
* https://github.com/circonus-labs/libmtev/pull/389/files#diff-9ddf6c97c258d08cbe3203e87cd44cedR147
* https://github.com/circonus-labs/libmtev/pull/389/files#diff-9ddf6c97c258d08cbe3203e87cd44cedR208
I don't see a pattern.

HeinrichHartmann:
LGTM

PR #392, opened 2018-04-27
Heap profiling

rileyberton:


src/examples/example1.c
```diff
@@ -69,7 +70,7 @@ parse_cli_args(int argc, char * const *argv) {
 static mtev_hook_return_t
 on_node_updated(void *closure, mtev_cluster_node_changes_t node_changes, mtev_cluster_node_t *updated_node, mtev_cluster_t *cluster,
     struct timeval old_boot_time) {
-  mtev_boolean i_am_oldest = mtev_cluster_am_i_oldest_node(my_cluster);
+  mtev_boolean i_am_oldest = mtev_cluster_am_i_oldest_visible_node(my_cluster);
```
postwait: how'd this get in here?

postwait: oh.. did this not even compile before?

rileyberton: It did not compile before.. whenever that change was introduced the examples were not built.

PR #393, opened 2018-05-08
Add option to disable JIT at runtime

cody271:
Use with environment variable `MTEV_JIT_OFF`.

For example: ```$ MTEV_JIT_OFF=1 /opt/circonus/bin/luamtev  <module>```

### DIFF
--- a/src/modules/lua.c
+++ b/src/modules/lua.c
@@ -843,6 +843,15 @@ package_manip_path(char *in, const char *find, const char *replace) {
   return npath;
 }
 
+static int MTEV_JIT_OFF(void) {
+  static int enabled = -1;
+  if(enabled == -1) {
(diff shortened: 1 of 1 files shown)


src/modules/lua.c
```diff
@@ -843,6 +843,15 @@ package_manip_path(char *in, const char *find, const char *replace) {
   return npath;
 }
 
+static int MTEV_JIT_OFF(void) {
+  static int enabled = -1;
```
postwait: enabled is a bad name as this is actually disabling JIT.

Can you change `enabled` to `jit_off`

cody271: Will do. (Agreed that is bad naming, it was blindly copied from `DEBUG_LOG_ENABLED` in `utils/mtev_log.c`..)

PR #399, opened 2018-05-16
Memmem

postwait:


src/utils/mtev_str.c
```diff
@@ -76,8 +75,25 @@ const char *strnstrn(const char *needle, int needle_len,
...
+                     const char *haystack, int haystack_len) {
+  return match_needle_haystack(needle, needle_len, haystack, haystack_len);
+}
+#endif
+
+void *
+mtev_memmem(const void *haystack, size_t haystack_len,
+            const void *needle, size_t needle_len) {
+#ifndef BROKEN_MEMMEM
+  return memmem(haystack, haystack_len, needle, needle_len);
```
rileyberton: This is not sufficient:

```In glibc 2.0, if needle is empty, memmem() returns a pointer to the
       last byte of haystack.  This is fixed in glibc 2.1.```

We need to check for `needle_len == 0` and return NULL.. or maybe return first byte of haystack?  what does searching for nothing mean?

Comment (postwait) [reply src/utils/mtev_str.c]: I think I'd prefer to check that in autoconf and just say BROKEN_MEMMEM... so what would a check for that look like?
```
memmem("abcd", 4", "a", 0) != NULL
```

Comment (rileyberton) [reply src/utils/mtev_str.c]: Nope, wrong.. glibc >2.1 does this:
```
  if (needle_len == 0)
    /* The first occurrence of the empty string is deemed to occur at
       the beginning of the string.  */
    return (void *) haystack;
```

Comment (rileyberton) [reply src/utils/mtev_str.c]: A valid check for brokeness would be:

```
const char *s = "abcd";
memmem(s, 4, "", 0) == (s + 3);
```

PR #403, opened 2018-05-18
Add support for configuration uints

postwait:


src/mtev_conf.c
```diff
@@ -2093,6 +2122,19 @@ mtev_conf_string_to_int32(const char *str) {
   return strtol(str, NULL, base);
 }
 
+/* No locks required */
+int
+mtev_conf_get_uint32(mtev_conf_section_t section,
+                     const char *path, uint32_t *value) {
+  char *str;
+  if(_mtev_conf_get_string(section,NULL,path,&str)) {
+    *value = (int)mtev_conf_string_to_uint32(str);
```
rileyberton: Why cast to `int` here?

postwait: C&P bug. fixed.

PR #407, opened 2018-06-11
More complete UUID function coverage.  Still no generate.

postwait:


src/utils/mtev_uuid_parse.c
```diff
@@ -72,3 +78,22 @@ void mtev_uuid_unparse_lower(const uuid_t uu, char *out)
   }
   *w = '\0';
 }
+
+void mtev_uuid_unparse(const uuid_t uu, char *out)
```
rileyberton: This function is fundamentally the same as the uuid_unparse_lower.  Recommend splitting this into a macro or inline function that we can call by passing in the correct `reverse_*ut` table.  Then have the individual functions call that macro/inline.

PR #410, opened 2018-06-12
deprecate atomics, use concurrencykit

postwait:


src/utils/mtev_log.c
```diff
@@ -539,12 +538,12 @@ static void *
 asynch_logio_writer(void *vls) {
   mtev_log_stream_t ls = vls;
   asynch_log_ctx *actx = ls->op_ctx;
-  int gen;
-  gen = mtev_atomic_inc32(&actx->gen);
+  uint32_t gen;
+  gen = ck_pr_faa_32(&actx->gen, 1) + 1;
```
sanjayar: Are you sure you want to faa( gen, 1) also add one? +1? Seems

src/utils/mtev_memory.c
```diff
@@ -685,8 +685,8 @@ fixed_umem_allocator_init(struct mtev_allocator *a, mtev_allocator_options_t opt
 mtev_allocator_t mtev_allocator_create(mtev_allocator_options_t opt) {
   mtev_allocator_t allocator = calloc(1, sizeof(*allocator));
   if(opt->name[0] == '\0') {
-    int id = mtev_atomic_inc32(&nallocators);
-    snprintf(opt->name, sizeof(opt->name), "mtev_umem_n%d_%d",
+    uint32_t id = ck_pr_faa_32(&nallocators, 1) + 1;
+    snprintf(opt->name, sizeof(opt->name), "mtev_umem_n%u_%d",
```
sanjayar: again not sure why the '+1' .

src/eventer/eventer.c
```diff
@@ -223,10 +225,7 @@ int64_t eventer_allocations_total(void) {
 }
```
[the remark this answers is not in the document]
postwait: preexisting interface... correct.

src/eventer/eventer_ports_impl.c
```diff
@@ -368,7 +368,7 @@ static int eventer_ports_impl_loop(int id) {
 
     ret = port_getn(spec->port_fd, pevents, MAX_PORT_EVENTS, &fd_cnt,
                     &__ports_sleeptime);
-    spec->wakeup_notify = 0; /* force unlock */
+    ck_spinlock_init(&spec->wakeup_notify); /* force unlock */
```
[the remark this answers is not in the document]
postwait: it is not unlocked...  we use it as a single-fire semaphore.... if we need to wake up the other thread we'll use a port_send, but that syscall would be wasteful if we were to wake it up more than once.  So we wrap it in a trylock (for a single fire behavior) and then once we're awake again, we just reinitialize it to an unlocked state for the next loop.

src/utils/mtev_log.c
postwait: mtev_atomic_inc32 did a cas-based add (of 1) and returned the new value.  So, yes?

PR #428, opened 2018-08-21
Granular control over lua_gc from module config.

postwait:
New defaults of step for 1000 iterations, then full.

 * gc_full: n = 0 no full, n > 0 full every n iterations
 * gc_step: arg to lua_gc LUA_GCSTEP
 * gc_stepmul: arg to lua_gc LUA_GCSETSTEPMUL
 * gc_pause: arg to lua_gc LUA_GCSETPAUSE (% growth)


src/modules/lua.c
```diff
@@ -58,6 +58,79 @@ static pthread_mutex_t mtev_lua_states_lock = PTHREAD_MUTEX_INITIALIZER;
...
+    p->pause_size = p->set_pause_size;
+  }
+
+  p->iters_since_full++;
+  if(p->full_every && (p->iters_since_full > p->full_every)) {
+    p->iters_since_full = 0;
+    lua_gc(L, LUA_GCCOLLECT, 0);
+    return;
+  }
+  lua_gc(L, LUA_GCSTEP, p->steps);
```
HeinrichHartmann: Check if `p->steps > 0` before calling lua_gc here?

src/modules/lua.c
```diff
@@ -58,6 +58,79 @@ static pthread_mutex_t mtev_lua_states_lock = PTHREAD_MUTEX_INITIALIZER;
 static mtev_hash_table mtev_coros;
 static pthread_mutex_t coro_lock = PTHREAD_MUTEX_INITIALIZER;
 
+struct lua_module_gc_params {
+  int iters_since_full;
+  int full_every;
+  int steps;
+  int steps_multiplier, set_steps_multiplier;
+  int pause_size, set_pause_size;
```
HeinrichHartmann: Your commit comment (below) is helpful to interpret the members of this struct. It would be helpful if it the comment would be somewhere in the source. Maybe here is a good place?
 
```
gc_full: n = 0 no full, n > 0 full every n iterations
gc_step: arg to lua_gc LUA_GCSTEP
gc_stepmul: arg to lua_gc LUA_GCSETSTEPMUL
gc_pause: arg to lua_gc LUA_GCSETPAUSE (% growth)
```

HeinrichHartmann:
See comments. LGTM otherwise.

src/modules/lua.c
postwait: No, we want to call with steps == 0, that's the default.

src/modules/lua.c
postwait: I strongly dislike docs in two places.  I've added the parameter documentation for the modules which gets pulled into the main published docs.

PR #451, opened 2018-10-02
Disable trace output

### LINKED ISSUE #450: Want option to disable trace output files
The `run_glider` function unconditionally saves the stdout of the configured glider to `tracedir`. Depending on the glider being used, this can end up being a rather large amount of output, and frequent crashes can cause the tracedir to grow quite large.

The operator should be able to disable this functionality if it is causing issues or provides insufficient value.

esproul:
Fixes #450

### DIFF
--- a/src/utils/mtev_watchdog.c
+++ b/src/utils/mtev_watchdog.c
@@ -56,6 +56,7 @@
 #endif
 
 #include "eventer/eventer.h"
+#include "mtev_conf.h"
 #include "mtev_log.h"
 #include "mtev_time.h"
 #include "mtev_watchdog.h"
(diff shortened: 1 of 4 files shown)


src/utils/mtev_watchdog.c
```diff
@@ -548,6 +555,7 @@ int mtev_watchdog_start_child(const char *app, int (*func)(void),
       sigaddset(&mysigs, SIGINT);
       setup_signals(&mysigs);
       mtev_monitored_child_pid = child_pid;
+      mtev_conf_get_boolean(MTEV_CONF_ROOT, "//watchdog/@save_trace_output", &save_trace_output);
```
postwait: we should not use mtev_conf from with utils/

We need to expose a function to set that boolean and call it from mtev_main.c dependent upon this config value

esproul: Addressed in latest commit.

PR #452, opened 2018-10-02
Add High Level Lua Api

HeinrichHartmann:
This PR introduces two classes `mtev.Proc` and `mtev.Api` which are extracted from similar scripts in  mtevbusted/lua-support.
These classes don't provide any new functionality, but they are rather wrappers around `mtev.HttpClient` and `mtev.spawn_process` that make common use cases more accessible.
In order not keep the documentation of the core functionality clean, I opted to create a new doc section: "High Level Lua API". (This is similar to the lua_* and luaL_* apis that lua itself provides.) Functions in the High Level API are marked as unstable for the time being.

Changes made along the way:
- Support for gitbook documentation embedded in lua code
- Add mtev.time()
- Add mtev.exec(), mtev.sh()
- Fix to mtev.waitfor()


HeinrichHartmann:
Replaced the internals of child.lua with Proc. Busted tests for libmtev and snowth are still running as expected, so I am reasonably confident that went fine.
Replacing the internals of api.lua with the new Api, is not worth it since the APIs are quite different, and there is just too little code to replace.

src/modules/lua-support/mtev/Api.lua
```diff
@@ -0,0 +1,188 @@
...
+--
+module("mtev.API", package.seeall)
+
+local HttpClient = require('mtev.HttpClient')
+
+local function HTTP(method, host, port, uri, headers, payload, sslconfig)
+  assert(method)
+  assert(host)
+  assert(port)
+  local ssl = not not sslconfig
```
cody271: `not not` is odd here .. how about `local ssl = sslconfig ~= nil` ?

src/modules/lua-support/mtev/Api.lua
```diff
@@ -0,0 +1,188 @@
...
+  if doc == nil then
+    error("json parse error: %s @%d\n", err, offset)
+  end
+  return doc:document()
+end
+
+--
+-- Class: API
+--
+Api = {};
```
cody271: nit: don't need the semicolons here and below

src/modules/lua-support/mtev/Proc.lua
```diff
@@ -0,0 +1,287 @@
...
+--\hlua term, status, errno = mtev.Proc:wait(timeout)
+--\brief wait for a process to terminate
+--\return trem is true if the process terminated and status, errno as returned by mtev.process:wait()
+--*/
+function Proc:wait(timeout)
+  if self.proc == nil then return nil end
+  local status, errno = self.proc:wait(timeout)
+  if status then
+    if mtev.WIFEXITED(status) or mtev.WIFSIGNALED(status) then
+      return true, status
```
cody271: Probably want to return `mtev.WEXITSTATUS(status)` here and below

src/modules/lua-support/mtev/Api.lua
HeinrichHartmann: `not not x` is an idiom from converting x to bool.

src/modules/lua-support/mtev/Proc.lua
HeinrichHartmann: Good point. Does that also work if WIFSIGNALED was true?

cody271: nope, from docs: 
```       
WEXITSTATUS(wstatus)
              returns the exit status of the  child.   This  consists  of  the
              least  significant  8 bits of the status argument that the child
              specified in a call to exit(3) or _exit(2) or  as  the  argument
              for a return statement in main().  This macro should be employed
              only if WIFEXITED returned true.
```

src/modules/lua-support/mtev/Api.lua
cody271: yes as discussed, this handles an accidental `false` here, all good

src/modules/lua-support/mtev/Proc.lua
HeinrichHartmann: Yeah, makes sense.

I found returning the status struct to be bad usability, so I was looking for something better.
Usually I am interested in: "Is the process dead or not?"
So, this is what I (tried?) to return in first return value. You also get the full status struct and error info in case you want to start digging.

Not the greatest API I have seen. Not sure how to improve it though.

PR #455, opened 2018-10-15
Default luamtev to an eventer concurrency of 1.

postwait:
Change the "concurrency" eventer config to support 0 for default
behavior.  Add a -n flag to luamtev to control concurrency and
default it to 1 if the flag is unused.


HeinrichHartmann:
LGTM

PR #458, opened 2018-10-29
Fix SMR in jobqs. make begin/end recursion safe

postwait:


src/utils/mtev_memory.c
```diff
@@ -72,6 +73,10 @@ void mtev_memory_init_thread(void) {
 }
 
 void mtev_memory_fini_thread(void) {
+  if(begin_end_depth > 1) {
+    mtev_memory_end();
+    begin_end_depth = 0; // setting this doesn't actually matter.
```
vasuraman11: not sure I understand the comment
begin_end_depth is decremented in mtev_memory_end() and then checked if it is equal to zero
and if zero, it calls ck_epoch_end(), closing the epoch
so it looks like if set to zero here, it will bypass the ck_epoch_end() call...

vasuraman11:
LGTM

PR #460, opened 2018-11-01
Make lua GC on next eventer tick.

postwait:
We can be in a lua call triggering into C completing and decide "we're done"
with the coroutine that is servicing the request.  From lua's perspective
we are done, but it is littered through the calling frames and is very likely
going to be used before we unwind.  So, instead of GC in place, we make
mtev_lua_gc schedule a GC on next tick which makes for a clean (empty) call
stack and greatly simplifies the next of lua/C call stacks.

### DIFF
--- a/src/modules/lua.c
+++ b/src/modules/lua.c
@@ -104,13 +104,14 @@ mtev_lua_set_gc_params(lua_module_closure_t *lmc, lua_module_gc_params_t *p) {
   }
 }
 
-void
-mtev_lua_gc(lua_module_closure_t *lmc) {
+static int
+mtev_lua_gc_callback(eventer_t e, int mask, void *c, struct timeval *now) {
+  lua_module_closure_t *lmc = c;
   lua_State *L = lmc->lua_state;
   lua_module_gc_params_t *p = lmc->gcparams;
   if(!p) {
     lua_gc(L, LUA_GCCOLLECT, 0);
-    return;
+    return 0;
   }
 
   if(p->steps_multiplier != p->set_steps_multiplier) {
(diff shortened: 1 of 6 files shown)


postwait:
This also uses the global registry instead of a "mtev_coros" global table for thread references.

src/modules/lua.c
```diff
@@ -183,11 +189,10 @@ mtev_lua_lmc_setL(lua_module_closure_t *lmc, lua_State *L) {
...
 void
 mtev_lua_cancel_coro(mtev_lua_resume_info_t *ci) {
-  lua_getglobal(ci->lmc->lua_state, "mtev_coros");
-  luaL_unref(ci->lmc->lua_state, -1, ci->coro_state_ref);
-  lua_pop(ci->lmc->lua_state, 1);
-  lua_gc(ci->lmc->lua_state, LUA_GCCOLLECT, 0);
   mtevL(nldeb, "coro_store <- %p\n", ci->coro_state);
+  luaL_unref(ci->lmc->lua_state, LUA_REGISTRYINDEX, ci->coro_state_ref);
+  lua_pop(ci->lmc->lua_state, 1);
+  mtev_lua_gc(ci->lmc);
```
cody271: I think we always want to do a full collection here ?
```suggestion
  lua_gc(ci->lmc->lua_state, LUA_GCCOLLECT, 0);
```

src/modules/lua.c
```diff
@@ -183,11 +189,10 @@ mtev_lua_lmc_setL(lua_module_closure_t *lmc, lua_State *L) {
 
 void
 mtev_lua_cancel_coro(mtev_lua_resume_info_t *ci) {
-  lua_getglobal(ci->lmc->lua_state, "mtev_coros");
-  luaL_unref(ci->lmc->lua_state, -1, ci->coro_state_ref);
-  lua_pop(ci->lmc->lua_state, 1);
-  lua_gc(ci->lmc->lua_state, LUA_GCCOLLECT, 0);
   mtevL(nldeb, "coro_store <- %p\n", ci->coro_state);
+  luaL_unref(ci->lmc->lua_state, LUA_REGISTRYINDEX, ci->coro_state_ref);
+  lua_pop(ci->lmc->lua_state, 1);
```
cody271: Isn't the `lua_pop()` for the `mtev_coros` table ? Don't think we need it anymore..
```suggestion
```

cody271:
> This also uses the global registry instead of a "mtev_coros" global table for thread references.

:+1:

postwait:
All nits addressed.

src/modules/lua.c
postwait: This would move it back inline, I added a mtev_lua_gc_full function that will force a full on next tick.

PR #462, opened 2018-11-03
Make ncct write operations thread safe

postwait:


src/mtev_console.c
```diff
@@ -79,20 +79,51 @@
...
 
+#define OL(a) do { \
+  pthread_mutex_lock(&(a)->outbuf_lock); \
+} while(0)
+#define OUL(a) do { \
+  pthread_mutex_unlock(&(a)->outbuf_lock); \
+} while(0)
+
+static int
+nc_write_with_lock(mtev_console_closure_t ncct, const void *buf, int len) {
```
HeinrichHartmann: Nit: Naming of this function is confusing, to me. We are not locking. We assume the lock is being hold by the caller.

src/mtev_console.c
```diff
@@ -269,12 +302,14 @@ mtev_console_continue_sending(mtev_console_closure_t ncct,
   eventer_t e = ncct->e;
   if(!ncct->outbuf_len) return 0;
   if(ncct->output_cooker) ncct->output_cooker(ncct);
+  OL(ncct);
```
HeinrichHartmann: Do we need to lock the access to outbuf_len two lines before this?
https://github.com/circonus-labs/libmtev/blob/5812034e0fef7d34c48deedc5cb4d4f65ebe76a8/src/mtev_console.c#L303

HeinrichHartmann:
Looks like accesses to outbuf_* are appropriately locked.

There are two places where it's not 100% certain. Otherwise looks fine.

src/mtev_console.c
postwait: It's a consistency-safe race.  Should be fine.

src/mtev_console.c
postwait: This is fairly consistent.  The caller call the _with_lock function with a lock.

PR #472, opened 2018-11-16
Fix thread migration for ACO rest calls.

postwait:
By default, if pools were specified, thread-fanout happened.  This
now always happens as "no pool" is treated correctly as the default
pool.

### DIFF
--- a/src/mtev_rest.c
+++ b/src/mtev_rest.c
@@ -305,6 +305,9 @@ mtev_http_get_websocket_handler(mtev_http_rest_closure_t *restc)
 
 static rest_request_handler
 mtev_http_get_handler(mtev_http_rest_closure_t *restc, mtev_boolean *migrate) {
+  /* This function is short-circuited by setting fastpath, which it does */
+  if(restc->fastpath) return restc->fastpath;
+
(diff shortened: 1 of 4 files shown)


src/mtev_rest.c
```diff
@@ -315,10 +318,17 @@ mtev_http_get_handler(mtev_http_rest_closure_t *restc, mtev_boolean *migrate) {
     restc->fastpath = rule->handler;
     restc->closure = rule->closure;
     restc->aco_enabled = rule->aco_enabled;
-    if(rule->pool) {
+    *migrate = mtev_false;
+    if(!mtev_http_session_aco(restc->http_ctx)) {
```
cody271:
```suggestion
    if(mtev_http_session_aco(restc->http_ctx)) {
```
Shouldn't this be testing for true not false here ..?

postwait: I don't think so. We can do migration until we go aco. After that, it is not okay. We should only do the migration in the case where we are not aco.

cody271: Ah, you're correct .. This is for before.

PR #476, opened 2018-12-04
Make epoll_wait use a smaller possible return set.

postwait:
This saves quite a bit of memory on highly concurrent configs.
This simply limits each epoll_wait to 1024 fds to attend to, any
subsequent ones will be picked up on the next iteration.

### DIFF
--- a/src/eventer/eventer_epoll_impl.c
+++ b/src/eventer/eventer_epoll_impl.c
@@ -456,9 +456,10 @@ static int eventer_epoll_impl_loop(int id) {
   (void)id;
   struct epoll_event *epev;
   struct epoll_spec *spec;
+  int max_fds_at_once = 1024;
 
   spec = eventer_get_spec_for_event(NULL);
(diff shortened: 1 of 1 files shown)


src/eventer/eventer_epoll_impl.c
```diff
@@ -456,9 +456,10 @@ static int eventer_epoll_impl_loop(int id) {
   (void)id;
   struct epoll_event *epev;
   struct epoll_spec *spec;
+  int max_fds_at_once = 1024;
```
bdunavant: maxfds is defined based on the particulars of the OS (if i'm reading eventer_impl_setrlimit correctly).  While probably unlikely anywhere our stuff is running, is it possible for an rlimit to be less than 1024?  And if so, would that break when you pass epoll_wait a value that large?

postwait: No risk. This just sets the limit of the size of output events produced by epoll_wait.

PR #478, opened 2018-12-07
Have the intern release perform compaction

postwait:


esproul:
Is this lightweight enough that high-churn use cases don't turn pathological?

PR #479, opened 2018-12-07
http2 support

postwait:


src/mtev_listener.c
```diff
@@ -665,6 +669,40 @@ mtev_listener_reconfig(const char *toplevel) {
...
+                         struct timeval *now) {
+  eventer_ssl_ctx_t *sslctx;
+  if((sslctx = eventer_get_eventer_ssl_ctx(e)) == NULL) return 0;
+  const uint8_t *alpn = NULL;
+  uint8_t alpnlen = 0;
+  eventer_ssl_get_alpn_selected(sslctx, &alpn, &alpnlen);
+  if(alpnlen && alpn) {
+    const char *name = eventer_name_for_callback(eventer_get_callback(e));
+    char buff[128];
+    snprintf(buff, sizeof(buff), "%s/alpn:%.*s", name, (int)alpnlen, alpn);
```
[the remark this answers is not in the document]
bdunavant: (Yes, it's using snprintf, but we don't check for truncation, so you could get collisions on names)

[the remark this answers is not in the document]
postwait: Anything but short sensible names will be unmanagable.  In this case, we'll fail to match and the listener won't function, which seems a desirable outcome.

src/eventer/eventer_SSL_fd_opset.c
```diff
@@ -848,6 +850,85 @@ eventer_ssl_ctx_new(eventer_ssl_orientation_t type,
...
+  if(npn) {
+    void *vf = NULL;
+    if(mtev_hash_retrieve(&alpn_funcs, npn, strlen(npn), &vf)) {
+      f = vf;
+    }
+  }
+  if(npn && f) {
+    ctx->npn = malloc(1 + strlen(npn));
+    ctx->npn[0] = strlen(npn);
+    memcpy(ctx->npn+1, npn, ctx->npn[0]);
```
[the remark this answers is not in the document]
postwait: Fixed.

src/eventer/eventer_SSL_fd_opset.c
```diff
@@ -848,6 +851,85 @@ eventer_ssl_ctx_new(eventer_ssl_orientation_t type,
...
 }
 
+#ifndef OPENSSL_NO_NEXTPROTONEG
+static int next_proto_cb(SSL *ssl, const unsigned char **data,
+                         unsigned int *len, void *arg) {
+  (void)ssl;
+  eventer_ssl_ctx_t *ctx = (eventer_ssl_ctx_t *)arg;
+
+  *data = ctx->npn;
+  *len = (unsigned int)(1 + *ctx->npn);
```
bincircon: how do we handle Len overflow.

src/mtev_http.c
```diff
@@ -335,274 +127,242 @@ struct bchain *bchain_from_data(const void *d, size_t size) {
...
-   default:
-    break;
+#ifdef va_copy
+  va_copy(copy, arg);
+  len = vsnprintf(buffer, sizeof(buffer), format, copy);
+  va_end(copy);
+#else
+  len = vsnprintf(buffer, sizeof(buffer), format, arg);
+#endif
+  if(len > (int)sizeof(buffer)) {
```
bincircon: should it be len >= (int) sizeof(buffer)
int vsnprintf (char * s, size_t n, const char * format, va_list arg );
Write formatted data from variable argument list to sized buffer
Composes a string with the same text that would be printed if format was used on printf, but using the elements in the variable argument list identified by arg instead of additional function arguments and storing the resulting content as a C string in the buffer pointed by s (taking n as the maximum buffer capacity to fill).

If the resulting string would be longer than n-1 characters, the remaining characters are discarded and not stored, but counted for the value returned by the function

src/eventer/eventer_SSL_fd_opset.c
postwait: I don't think it can overflow. ctx->npn is a uint8_t, so [0, 255] + 1 is [1,256] which fits in *len as an unsigned int.

src/mtev_http.c
postwait: addressed.

PR #480, opened 2018-12-10
Support decoding 'URL' Alphabet base64 messages (rfc4648)

postwait:


cody271:
Confirmed, this matches ```Table 2: The "URL and Filename safe" Base 64 Alphabet``` from [RFC 4648](https://tools.ietf.org/rfc/rfc4648.txt).

Also known as [`"base64url"`](https://en.wikipedia.org/wiki/Base64#URL_applications) format.

PR #488, opened 2019-02-06
Make mtev_skiplist Safer If Misused

pamaddox:
There are many functions in mtev_skiplist that will crash if passed a
null skiplist. Rather than crash, we should check for this and return a
failure.

### DIFF
--- a/src/utils/mtev_skiplist.c
+++ b/src/utils/mtev_skiplist.c
@@ -225,8 +225,10 @@ static int mtev_skiplisti_find_compare(mtev_skiplist *sl,
   int count=0;
 
   /* If we weren't passed a skiplist, there's nothing
-   * to compare */
-  if (!sl) return 0;
+   * to compare - we don't want to return 0, since that
+   * conveys equality - there's no good way out of this,
+   * so we should just assert */
+  mtevAssert(sl != NULL);
 
   if(ret) *ret = NULL;
   if(prev) *prev = NULL;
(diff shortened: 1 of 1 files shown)


src/utils/mtev_skiplist.c
```diff
@@ -217,6 +223,11 @@ static int mtev_skiplisti_find_compare(mtev_skiplist *sl,
                                        mtev_skiplist_comparator_t comp) {
   mtev_skiplist_node *m = NULL;
   int count=0;
+
+  /* If we weren't passed a skiplist, there's nothing
+   * to compare */
+  if (!sl) return 0;
+
```
postwait: zero means the things are equal... this should actually still crash.

src/utils/mtev_skiplist.c
```diff
@@ -267,6 +278,7 @@ void *mtev_skiplist_previous(mtev_skiplist *sl, mtev_skiplist_node **iter) {
   return (*iter)?((*iter)->data):NULL;
 }
 void *mtev_skiplist_data(mtev_skiplist_node *m) {
+  if (!m) return NULL;
```
postwait: I think this should crash

src/utils/mtev_skiplist.c
```diff
@@ -361,18 +375,20 @@ mtev_skiplist_node *mtev_skiplist_insert_compare(mtev_skiplist *sl,
 }
 int mtev_skiplist_remove(mtev_skiplist *sl,
                          const void *data, mtev_freefunc_t myfree) {
+  if(!sl) return 0;
   if(!sl->compare) return 0;
   return mtev_skiplist_remove_compare(sl, data, myfree, sl->comparek);
 }
 int mtev_skiplist_remove_node(mtev_skiplist *sl, mtev_skiplist_node *m,
                               mtev_freefunc_t myfree) {
+  if (!m) return 0;
```
postwait: This should crash too.

pamaddox: Addressed

src/utils/mtev_skiplist.c
pamaddox: Addressed

src/utils/mtev_skiplist.c
pamaddox: Addressed

pamaddox:
Addressed requested changes

PR #490, opened 2019-02-12
Support stats exposure for mtev_intern pools

postwait:


esproul:
Still LGTM, tested the new list command, works great.

PR #491, opened 2019-02-14
make mtev_hash__hash be XXH64 for speed

postwait:


cody271:
We should probably update our copy of `xxHash` to the latest while we're at it..

PR #492, opened 2019-02-14
Cleanup some intern code.

postwait:
This should yield no functional change.  It adds some asserts and
macros for commonly repeated ops and reworks a few predicated to
avoid unsigned rolling (that should never have been possible anyway).
Make the test a bit more rigorous.

### DIFF
--- a/src/utils/mtev_intern.c
+++ b/src/utils/mtev_intern.c
@@ -162,6 +162,8 @@ typedef struct mtev_intern_internal {
   uint8_t   v[0];
 } mtev_intern_internal_t;
 
+#define WRAPPED_SIZE(l) (offsetof(mtev_intern_internal_t, v) + (((l) + 3) & ~3))
+
 struct mtev_intern_pool_extent {
   size_t    size;
   void     *base;
@@ -314,6 +316,7 @@ borrow_free_node(mtev_intern_pool_t *pool,
     mtev_plock_drop_r(&l->lock);
     return NULL;
   }
+  assert(n->size >= l->lsize); // this node belongs here
   assert(len > 8);        // header
   assert((len & 3) == 0); // aligned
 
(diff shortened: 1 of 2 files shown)


test/intern_test.c
```diff
@@ -83,7 +101,13 @@ void load_words(void) {
   while(fgets(buff, sizeof(buff), fp)) {
     if(strlen(buff) < 1) continue;
     if(buff[strlen(buff)-1] == '\n') buff[strlen(buff)-1] = '\0';
-    words[word_cnt++] = strdup(buff);
+    int upsize = mtev_rand() % 4096;
+    if(upsize < strlen(buff)) upsize = strlen(buff);
+    char *str = malloc(upsize+1);
```
cody271: Where do we call `free(str)`?

src/utils/mtev_intern.c
```diff
@@ -322,7 +325,8 @@ borrow_free_node(mtev_intern_pool_t *pool,
   /* If we can manage to borrow the requested len without needing
    * to move this node into another freeslot, we can just "borrow"
    * the bytes from be done. */
-  while((oldsize = ck_pr_load_64(&n->size)) - len >= l->lsize) {
+  while((oldsize = ck_pr_load_64(&n->size)) >= l->lsize + len) {
```
cody271: I don't quite follow the changes here and below.. Why is `size >= lsize + len` better than `size - len >= lsize` ?

test/intern_test.c
postwait: It is not freed, it's building a table that's used until exit.

src/utils/mtev_intern.c
postwait: they are unsigned types.  So A - B >= C could have very bad behavior if  B > A.. (underflow).

By adding, you guarantee that you'd have to overflow to break (which is impracticle given that lengths of allocations here are stored in a unsigned:23

PR #493, opened 2019-02-18
Intern unswole

postwait:


src/utils/mtev_intern.c
```diff
@@ -245,19 +249,18 @@ return_free_node(mtev_intern_pool_t *pool,
...
    * node than we have segments... cap it.
    */
   if(idx >= pool->nfreeslots) idx = pool->nfreeslots - 1;
 
-  mtev_plock_take_s(&pool->freeslots[idx].lock);
-  node->next = pool->freeslots[idx].head;
-  pool->freeslots[idx].head = node;
+  do {
+    node->next = ck_pr_load_ptr(&pool->freeslots[idx].head);
+  } while(!ck_pr_cas_ptr(&pool->freeslots[idx].head, node->next, node));
```
bdunavant: I'm not sure I understand this.   We set `node->next = head`.  We then atomically check ([ck_pr_cas_ptr](http://concurrencykit.org/doc/ck_pr_cas.html)) if `node->next == head`, and if it's not then we assign it to node (new_value) return true, negate that and exit the loop.  Otherwise, we do nothing (!false) and loop setting node->next = head again.  It seems to me, that unless another thread is acting on these values, that it will infinite loop.

bdunavant: err I flipped that in my head.  If it's == then it assigns to node and we exit.  But I don't understand why we don't just assign it to node in the first place.  What does this buy us that the seek lock did not?

bdunavant: Oh.  It lets us do the same two operations, without a seek lock.  And if something else changed the list under us, then it won't match, so we do it over.  Okay, nevermind.

PR #494, opened 2019-02-19
Fix intern race on freelist head acquisition

postwait:


src/utils/mtev_intern.c
```diff
@@ -614,28 +615,38 @@ mtev_intern_internal_t *mtev_intern_pool_find(mtev_intern_pool_t *pool, size_t l
...
-      /* The first time we can't find an allocation,
-       * we should attempt to unstage free nodes if there are any
-       * and if that is successful, try again.
-       */
+
+    /* If we can't find an allocation,
+     * we should attempt to unstage free nodes if there are any
+     * and if that is successful, try again.
+     */
+    if(attempt < 0) {
```
bdunavant: attempt is initialized at 1, and the only modification to it is attempt++.  I don't think this can ever occur?

src/utils/mtev_intern.c
```diff
@@ -614,28 +615,38 @@ mtev_intern_internal_t *mtev_intern_pool_find(mtev_intern_pool_t *pool, size_t l
...
-        minsize += pool->freeslots[i].lsize * pool->freeslots[i].cnt;
-      }
+    }
+    /* We think we have no free space, but we might want to just do a compaction
+     * right here and now. */
+    uint64_t total_size = 0;
+    for(struct mtev_intern_pool_extent *node = pool->extents; node; node = node->next) {
+    if(!node->internal) {
+      total_size += node->size;
+    }
```
bdunavant: nit: indentation

src/utils/mtev_intern.c
postwait: Correct. elided.

PR #502, opened 2019-03-27
check return AMPQ name

bincircon:


bincircon:
@jesus, in webex, there is assert call watchdog
mtevAssert(queuename.bytes != NULL); after call amqp_queue_declare

I dig over AMPQ, i am wonder if we pass queuename to below call, will the r->queue NULL because queue is not assigned internally by AMQP. 
amqp_queue_declare_ok_t *r =
        amqp_queue_declare(conn->conn, 1,
                           conn->queuename ? amqp_cstring_bytes(conn->queuename) : amqp_empty_bytes, 0, 0, 0, 1,
                           amqp_empty_table);
      die_on_amqp_error(amqp_get_rpc_reply(conn->conn), "declaring queue");

src/modules/mtev_amqp.c
```diff
@@ -355,7 +355,11 @@ rabbitmq_manage_connection(void *vconn) {
                            conn->queuename ? amqp_cstring_bytes(conn->queuename) : amqp_empty_bytes, 0, 0, 0, 1,
                            amqp_empty_table);
       die_on_amqp_error(amqp_get_rpc_reply(conn->conn), "declaring queue");
-      queuename = amqp_bytes_malloc_dup(r->queue);
```
postwait: Ths amqp_bytes_malloc_dup function unconditionally calls malloc.  The only reason we could get a null queuename is if that malloc returned NULL.  Is the idea that the r->queue.len == 0?  Usually `malloc(0)` does not return NULL.

postwait: I don't understand how amqp_queue_declare could succeed and not have a queue name.

bincircon:
@postwait please review, i noticed that static void *
rabbitmq_manage_connection(void *vconn) {
  struct timeval fiveseconds = { 5, 0 };
  struct amqp_conn *conn = vconn;

  while(1) {
it is doing a while for no block socket, create and destroy and recreate again, is it intentional.

postwait:
Yes, that blocking while(1) is intentional.  That is run in a dedicated thread.

bincircon:
@postwait it add some safe check, can we merge this

PR #504, opened 2019-03-28
Suppress GCC 7 warnings for "-Wimplicit-fallthrough" and "-Wclobbered"

cody271:
Fixes `libmtev` build with GCC 7 on newer Ubuntu. (Follow up to PR #370)

### DIFF
--- a/src/eventer/eventer_jobq.c
+++ b/src/eventer/eventer_jobq.c
@@ -695,6 +695,8 @@ jobq_thread_should_terminate(eventer_jobq_t *jobq, mtev_boolean want_reduce) {
   }
   return mtev_false;
 }
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wclobbered"
 void *
 eventer_jobq_consumer(eventer_jobq_t *jobq) {
   eventer_job_t *job;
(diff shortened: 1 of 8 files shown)


postwait:
Another option is to fix the code. What would that look like?

cody271:
For `-Wimplicit-fallthrough`, it looks these are all valid uses of fall through in a switch statement.

Personally I think warning suppression with `#pragma` is much clearer than using the special comment `/* fall through */`. Though we actually are already using that [here](https://github.com/circonus-labs/libmtev/blob/master/src/utils/mtev_watchdog.c#L735).. thoughts?

cody271:
For `-Wclobbered`, there we could be an actual bug there. I don't honestly understand the code in [`eventer_jobq_consumer()`](https://github.com/circonus-labs/libmtev/blob/master/src/eventer/eventer_jobq.c#L697) enough to know for sure. 

Documentation for that warning recommends adding `volatile`, but that shouldn't be necessary here and actually conflicts with `libck` primitives.

postwait:
I think the advantage of the `/* fall through */` comment is that is must be specified at use (per-case) whereas the pragma tends to wrap the whole switch.  I think we should comment.

postwait:
As for the clobber, since we use ck all over the place and this is the only warning site, we should figure out what's going on b/c the warning could be a real indicator of a flaw.

PR #511, opened 2019-04-11
Add statistics for lua in-VM runtime and GC operations.

postwait:


src/modules/lua.c
```diff
@@ -57,6 +59,8 @@ static mtev_hash_table mtev_lua_states;
 static pthread_mutex_t mtev_lua_states_lock = PTHREAD_MUTEX_INITIALIZER;
 static mtev_hash_table mtev_coros;
 static pthread_mutex_t coro_lock = PTHREAD_MUTEX_INITIALIZER;
+static stats_ns_t *lua_stats_ns;
+static stats_handle_t *gc_total, *gc_full, *gc_latency;
```
cody271: We could also report `LUA_GCCOUNT` here fairly easily.

src/modules/lua_general.c
```diff
@@ -810,7 +839,9 @@ mtev_console_lua_repl_execute(mtev_console_closure_t ncct,
   lua_remove(L, -2);  /* pop extras */
   lua_pushstring(L, buff);
   lua_pushConsole(L, ncct);
-  rv = lua_pcall(L, 2, LUA_MULTRET, -4);
+  VM_TIME_BEGIN
+  rv = lua_pcall(L, 3, LUA_MULTRET, -4);
```
cody271: Why are we changing `2` to `3` here ??

postwait: OUCH. Editor accident

src/modules/lua.c
postwait: it's a bit challenging to do it across all the VMs

PR #512, opened 2019-04-12
Cov work

postwait:


src/eventer/eventer_SSL_fd_opset.c
```diff
@@ -1197,7 +1197,7 @@ eventer_SSL_close(int fd, int *mask, void *closure) {
   } else {
     rv = close(fd);
   }
-  if(mask) *mask = 0;
+  *mask = 0;
```
pamaddox: Do we know for a fact that `mask` isn't NULL here? If no, seems like this could introduce a crash bug.

postwait: We use it up above as if it is not-NULL, so yes.

PR #513, opened 2019-04-20
Fix failure to AC_CHECK_LIB ck_ht_init on darwin

redhotpenguin:
My autoconf foo is not so great, but this change resolves the following
build error on darwin by adding -L to the full libpath, which finds libck (installed
either via source or homebrew")

error - "configure: error: *** concurrencyckit (libck) is required ***"


redhotpenguin:
Closing this one; will open a new PR in a bit, working through a couple other build issues on Darwin

PR #520, opened 2019-05-24
Address REST ACL hole

postwait:


postwait:
There is a lot of debugging added to help troubleshoot pattern issues in the future, but the gist of the fix is to move the `->auth(...)` application above the migrate logic in handler selection.

ChangeLog.md
```diff
@@ -4,6 +4,9 @@
 
 ## 1.6
 
+ * SECURITY: fix but in http authentication handling where thread fan out
```
esproul: "bug"

PR #526, opened 2019-06-12
Lua socket close yield (2)

HeinrichHartmann:
Avoid deadlock caused by eventer_t:close() in lua

Variant of https://github.com/circonus-labs/libmtev/pull/525

### DIFF
--- a/src/modules/lua_mtev.c
+++ b/src/modules/lua_mtev.c
@@ -224,6 +224,21 @@ inbuff_addlstring(struct nl_slcl *cl, const char *b, int l) {
 }
 
 static int
+nl_resume(eventer_t e, int mask, void *vcl, struct timeval *now) {
+    (void) mask;
+    (void) now;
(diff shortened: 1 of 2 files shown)


src/modules/lua_mtev.c
```diff
@@ -223,6 +223,21 @@ inbuff_addlstring(struct nl_slcl *cl, const char *b, int l) {
   cl->inbuff_len += l;
 }
 
+static int
+nl_resume(eventer_t e, int mask, void *vcl, struct timeval *now) {
+    (void) mask;
+    (void) now;
+    lua_State *L = vcl;
+    mtev_lua_resume_info_t *ci = mtev_lua_find_resume_info(L, mtev_false);
```
HeinrichHartmann: We can't directly pass the resume_info* as vcl, right?

HeinrichHartmann:
rebased.

PR #528, opened 2019-06-13
Add timeout parameter to lua/socket:connect()

HeinrichHartmann:
This is modeled after the lua_mtev_socket_read function.

### DIFF
--- a/src/modules/lua_mtev.c
+++ b/src/modules/lua_mtev.c
@@ -968,15 +968,27 @@ socket_connect_timeout(eventer_t e, int mask, void *closure, struct timeval *now
   mtev_lua_resume_info_t *ci;
   lua_timeout_callback_ref* cb_ref;
   lua_State *L;
+  struct nl_slcl *cl;
 
   cb_ref = (lua_timeout_callback_ref*) closure;
   L = cb_ref->L;
   ci = mtev_lua_find_resume_info(L, mtev_true);
   assert(ci);
 
-  // remove the original event
+  // Remove the original event from the event loop, and close the socket.
   eventer_remove_fde(cb_ref->timed_out_eventer);
   mtev_lua_deregister_event(ci, cb_ref->timed_out_eventer, 0);
+
+  // When we return to the event loop `timed_out_eventer` will be freeed.
+  // Make a copy so that the lua socket object has something to work with:
+  cl = eventer_get_closure(cb_ref->timed_out_eventer);
+  *(cl->eptr) = eventer_alloc_copy(cb_ref->timed_out_eventer);
+  mtev_lua_register_event(ci, *cl->eptr);
+
+  // The socket will be closed when either
+  // (a) e:close() is called explicitly on the lua object (recommended),
+  // (b) the lua object is GCed.
+  // No need to explicitly close it here.
 
   // return into the original Lua call which spawned this timeout
   lua_pushinteger(L, -1);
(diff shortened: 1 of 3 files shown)


src/modules/lua_mtev.c
```diff
@@ -953,9 +961,38 @@ static int
...
+
+  cb_ref = (lua_timeout_callback_ref*) closure;
+  L = cb_ref->L;
+  ci = mtev_lua_find_resume_info(L, mtev_true);
+  assert(ci);
+
+  // remove the original event
+  eventer_remove_fde(cb_ref->timed_out_eventer);
+  mtev_lua_deregister_event(ci, cb_ref->timed_out_eventer, 0);
+
```
HeinrichHartmann: There are two additional lines in the socket_read_timeout function, that I don't understand, and were thus not copied over to our timeout handler:
```
*(cl->eptr) = eventer_alloc_copy(cb_ref->timed_out_eventer);
mtev_lua_register_event(ci, *cl->eptr);
```
For me it looks like we are re-scheduling the timeout event. Which makes no sense to me.

src/modules/lua_mtev.c
```diff
@@ -953,9 +961,38 @@ static int
...
+  // return into the original Lua call which spawned this timeout
+  lua_pushinteger(L, -1);
+  lua_pushstring(L, "timeout");
+  ci->lmc->resume(ci, 2);
```
HeinrichHartmann: socket:connect() returns `-1, "timeout"` on timeout. socket:read() returns nil. Hence the need for a separate function.

src/modules/lua_mtev.c
postwait: We have to track the event, otherwise it will leak.. It is no longer tracked in the eventer (eventer_remove_fde)... so  e will be freed when we return and we can no longer keep a reference to it.  So we make a copy of it and put that in our lua state (mtev_lua_register_event).  This way, when the coro finishes, we know that we can close that event out (fd).

postwait: Where does the "timed_out_eventer" get closed?

src/modules/lua_mtev.c
```diff
@@ -1029,6 +1070,21 @@ mtev_lua_socket_connect(lua_State *L) {
   }
   if(rv == -1 && errno == EINPROGRESS) {
     /* Need completion */
+    /* setup timeout handler */
+    if (lua_gettop(L) >= 4 && lua_isnumber(L, 4)) {
+      double timeout_user = lua_tonumber(L, 4);
+      int timeout_s = floor(timeout_user);
+      int timeout_us = (timeout_user - timeout_s) * 1000000;
+      lua_timeout_callback_ref* cb_ref = malloc(sizeof(lua_timeout_callback_ref));
```
postwait: use `calloc`, it's good practice.

src/modules/lua_mtev.c
HeinrichHartmann: I see. I'll change the code to make a copy of the event and update cl->eptr.

> Where does the "timed_out_eventer" get closed?
It will get closed when we GC the original eventer object (the one created by mtev.socket()).
It might be better to close it here, though.

HeinrichHartmann: opted against the automatic close. With C/POSIX sockets you also have to call close explicitly.

src/modules/lua_mtev.c
HeinrichHartmann: done

HeinrichHartmann:
This appears to work. I don't see lingering sockets after timeout events in my testing. (As reported by `sudo ss -n -p | grep noit` on centos7).

PR #532, opened 2019-06-24
A Semaphore implementation for lua_mtev

HeinrichHartmann:


src/modules/lua-support/mtev/extras.lua
```diff
@@ -223,5 +223,31 @@ function mtev_coros_resume(co,...)
...
+--\brief returns true of the semaphore lock could be acquired within the given timeout, false if not.
+--
+--At the time of this writing, the implementation of this functions uses polling.
+--Expect delays between calls to :release() and subsequent :acquire() returning.
+--*/
+--
+-- Comment: The polling version of semaphore:acquire() is much easier to implement in lua than in C,
+-- since, we can't continue executing a C function after an mtev.sleep(). Ultimately we want this
+-- function to suspend the coro and wake up when :release() is called by another thread/coro.
+function _G.mtev.__semaphore_acquire(self, timeout, poll_interval)
```
postwait: Is timeout a floating point number of seconds?  If not, maybe it should be otherwise it will need to be documented.

HeinrichHartmann: yes

PR #533, opened 2019-06-24
Add module/lua/xcall.json

HeinrichHartmann:
This is an initial version. There are some obvious improvements that can be done in the future:
- Use lua_web to handle xcall requests
- Make xcall a lua function that can be used from lua_web handlers

With this PR we only aim for the immediate use case.
The implementation borrows heavily from /module/lua/stats.json

### DIFF
--- a/src/modules/lua.c
+++ b/src/modules/lua.c
@@ -718,6 +718,14 @@ mtev_lua_xcall_reporter(eventer_t e, int mask, void *closure,
   return 0;
 }
 
+/* GET /module/lua/xcall.json
+
+   Calls _G.xcall(querystring_table) in each Lua State.
+   Returns serialized output values as as JSON object of the form:
+
+        { "$stateptr" : <val> }
+
+*/
 static int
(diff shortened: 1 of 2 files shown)


src/modules/lua.c
```diff
@@ -664,6 +665,88 @@ mtev_console_show_lua(mtev_console_closure_t ncct,
...
+      lua_getglobal(L, "mtev");
+      lua_getfield(L, -1, "tojson");
+      lua_remove(L, -2);
+      lua_insert(L, -2);
+      lua_call(L, 1, 1);
+      lua_getfield(L, -1, "unwrap");
+      lua_insert(L, -2);
+      lua_call(L, 1, 1);
+      mtev_json_object **udata = luaL_checkudata(L, -1, "mtev.json_object");
+      out = (*udata);
```
HeinrichHartmann: Not sure about the ownership of that json object here. Do we need to explicitly free/unref it after use?

- The `:uwrap()` function calls `mtev_json_get()` so the object survives the lua function's return.
- mtev_rest_show_lua_complete calls `mtev_json_put()` on the root object, which should (?) unref this value here as well.

PR #534, opened 2019-06-25
 Add Lpreloads config for lua_web/lua_general

HeinrichHartmann:
This gives us a place to register xcall handlers in lua_web states.


postwait:
I'm fine with the variable in the code being Lpreloads, but I think as for docs, changelog and user-facing name, it should just be called `preloads`.. it's a lua module, so it would make sense that the preloads were lua.

HeinrichHartmann:
This was done in analogy to `LPATH` and `CPATH` variables used by lua. It's your call. Happy to change it to "preloads"

PR #536, opened 2019-06-25
Fixes to ACO and hash APIs

postwait:


src/eventer/eventer_aco_opset.c
```diff
@@ -117,15 +117,15 @@ eventer_aco_init(void) {
 eventer_aco_t
 eventer_set_eventer_aco_co(eventer_t e, aco_t *co) {
   if(co == NULL) {
-    mtevAssert(e->opset == eventer_aco_fd_opset);
+    mtevAssert((uintptr_t)e->opset == (uintptr_t)eventer_aco_fd_opset);
```
bdunavant: I am assuming you are casting these for a purpose.  However, this same cast does not happen on lines 72, 89, or 480, or in the assignment on 135.  If it's important here, should it not also be important there?

src/eventer/eventer_aco_opset.c
```diff
@@ -162,14 +166,47 @@ eventer_aco_mode_asynch_wrapper(eventer_t e, int mask, void *closure, struct tim
     simple_ctx->func_ops(EVENTER_ASYNCH_CLEANUP, simple_ctx->closure);
   }
   else if(mask == EVENTER_ASYNCH_COMPLETE) {
+    bool zero;
     aco_t *co = simple_ctx->co;
     simple_ctx->func_ops(EVENTER_ASYNCH_COMPLETE, simple_ctx->closure);
+    ck_pr_dec_32_zero(&simple_ctx->gate->ref_cnt, &zero);
     free(simple_ctx);
```
bdunavant: You free simple_ctx here, but ctx->gate is still allocated, do not not need to free that as well, or does that get handled in eventer_aco_gate_wait()?  If so, you should probably comment that here

src/eventer/eventer_aco_opset.c
```diff
@@ -193,19 +231,34 @@ eventer_aco_simple_asynch_wrapper(eventer_t e, int mask, void *closure, struct t
     simple_ctx->func(simple_ctx->closure);
   }
   if(mask == EVENTER_ASYNCH) {
+    bool zero;
     aco_t *co = simple_ctx->co;
+    ck_pr_dec_32_zero(&simple_ctx->gate->ref_cnt, &zero);
     free(simple_ctx);
```
bdunavant: same.  free(simple_ctx) but not gate, i think since this is the ASYNCH path it just needs a comment

postwait:
nits addressed.

PR #542, opened 2019-07-10
Protect intern release from concurrent double-free

vasuraman11:
Another case where we might not see a caller misbehavior in context  - this will not allow it to slide (previously we would have decremented a zero refcnt to 0xFFFFFFFF and essentially ignored the double-release instead of asserting on it and getting a callstack right away)


vasuraman11:
ok, sounds good, since it is going to assert anyway, can actually change the copy that way too to reduce perf hit

PR #547, opened 2019-07-12
HTTP aco memory issues

postwait:


src/eventer/eventer_epoll_impl.c
```diff
@@ -448,8 +453,9 @@ static void eventer_epoll_impl_trigger(eventer_t e, int mask) {
       }
       master_fds[fd].e = NULL;
     }
-    eventer_free(e);
+    eventer_deref(e);
   }
+  eventer_deref(e);
```
bdunavant: Won't the fall-through from line 456 cause this to deref twice?  This is in a few other places too.

src/examples/aco.c
```diff
@@ -69,14 +69,18 @@ parse_cli_args(int argc, char * const *argv) {
...
 void subcall2(mtev_http_rest_closure_t *restc) {
   eventer_aco_gate_t gate = eventer_aco_gate();
   mtevL(mtev_error, "subcall2 entry\n");
-  eventer_aco_simple_asynch_gated(gate, asynch_hello, restc);
-  eventer_aco_simple_asynch_gated(gate, asynch_hello, restc);
+  int count = 2;
+  const char *count_str = mtev_http_request_querystring(mtev_http_session_request(restc->http_ctx), "count");
+  if(count_str) count = atoi(count_str);
+  for(int i=0; i<count; i++) {
+    eventer_aco_simple_asynch_gated(gate, asynch_hello, restc);
```
bdunavant: If for some reason count is invalid, is it bad if this runs 0 times?

postwait: Count=0 is one of my test cases.. It works as expected.

src/eventer/eventer_epoll_impl.c
postwait: Correct. Intentional... The first one is the deref b/c the event is no longer used.  The second one is the deref matching the ref we did in the beginning of the function.

PR #549, opened 2019-07-16
Forcefully remove migrating thread from epoll.

postwait:
When a thread returns from an FD-triggered callback and the owner
has changed, we should always remove the thread from epoll because
if we do not, it will fail to add to the other loop spec and then
subsequently fire on the wrong thread and chaos will ensue.

### DIFF
--- a/src/eventer/eventer_epoll_impl.c
+++ b/src/eventer/eventer_epoll_impl.c
@@ -324,6 +324,7 @@ static void eventer_epoll_impl_trigger(eventer_t e, int mask) {
       eventer_deref(e);
       return;
     }
+    /* We've acquired the lock, recheck out predicate */
     if(master_fds[fd].e == NULL) { 
     /*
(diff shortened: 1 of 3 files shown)


src/eventer/eventer_epoll_impl.c
```diff
@@ -324,6 +324,7 @@ static void eventer_epoll_impl_trigger(eventer_t e, int mask) {
       eventer_deref(e);
       return;
     }
+    if(master_fds[fd].e == NULL) {
```
bdunavant: Line 312 just checked this above.  Won't this always be false?  If not, this likely needs a comment explaining why not after only a single if-branch has executed.

postwait: No. we have the lock now, it could have changed in between.

bdunavant: Sorry, indentation got me.

PR #550, opened 2019-07-17
Omnios aco cleanup

postwait:


src/eventer/eventer_impl.c
```diff
@@ -718,8 +718,9 @@ static int periodic_jobq_maintenance(eventer_t e, int mask, void *vjobq, struct
   (void)now;
```
vasuraman11: don't need this any longer because `now` is being used by the new code

PR #551, opened 2019-07-18
lua-mtev: Add mtev.getaddrinfo() function

HeinrichHartmann:


HeinrichHartmann:
As discussed in Slack: Keeping this as is, since it is idiomatic for lua.

PR #552, opened 2019-07-19
lua: Add mtev.inet_pton() function

HeinrichHartmann:
Apparently inet_pton is the best way to validate IP v4/v6 address: https://stackoverflow.com/a/4017219/1209380

Avoids ugly regexpes.


HeinrichHartmann:
I should have added a comment about this: This function can be trivially extended to also return the `struct in_addr` if there is ever the need for it.
So it would look like:
```
rc, fam, addr = mtev.inet_pton(addr_str)
```

postwait:
If it is trivial then extend, if it is not, then rename it.

HeinrichHartmann:
updated

src/modules/lua_mtev.c
```diff
@@ -5513,6 +5513,44 @@ nl_getaddrinfo(lua_State *L) {
...
+    lua_pushstring(L, "inet");
+    struct in_addr *u = lua_newuserdata(L, sizeof(struct in_addr));
+    *u = addr;
+    return 3;
+  }
+  rc = inet_pton(AF_INET6, host, &addr);
+  if (rc == 1) {
+    lua_pushboolean(L, 1);
+    lua_pushstring(L, "inet6");
+    struct in_addr *u = lua_newuserdata(L, sizeof(struct in_addr));
```
postwait: this isn't an in_addr... it's an `struct in6_addr`

HeinrichHartmann: Uh. The manual was not exactly transparent about this:

>     The inet_pton() function converts a presentation format address (that is,
>     printable form as held in a character string) to network format (**usually
>     a struct in_addr** or some other internal binary representation, in network
>     byte order).

Where is the return type documented?

PR #555, opened 2019-07-24
Make HTTP drain client data before request cleanup

postwait:


bdunavant:
Linking this to CIRC-3834

PR #557, opened 2019-07-24
Fix race in asynch event_close.

postwait:
Make the actual POSIX shutdown and close operations asynchronous
and not the whole event_close path b/c the event_close (and opset
dispatch) can happen while an event is being taken into or out of
ACO mode and it will abort.  This should be safe.


pamaddox:
LGTM

PR #559, opened 2019-08-05
Fix http1 chunked upload

HeinrichHartmann:


src/mtev_http1.c
```diff
@@ -835,12 +837,15 @@ mtev_http1_complete_request(mtev_http1_session_ctx *ctx, int mask, struct timeva
...
+    mtevL(http_debug, "[fd=%d] mtev_http -> read() = %d\n", CTXFD(ctx), len);
     if(len > 0)
-      mtevL(http_io, " mtev_http:read(%d) => %d [\n%.*s\n]\n", CTXFD(ctx), len, len, in->buff + in->start + in->size);
+      mtevL(http_io, "[fd=%d] mtev_http:read() => %d [\n%.*s\n]\n", CTXFD(ctx), len, len, in->buff + in->start + in->size);
     else
-      mtevL(http_io, " mtev_http:read(%d) => %d\n", CTXFD(ctx), len);
-    if(len == -1 && errno == EAGAIN) return mask;
+      mtevL(http_io, "[fd=%d] mtev_http:read() => %d\n", CTXFD(ctx), len);
+    if(len == -1 && errno == EAGAIN) {
+      sleep(1);
```
postwait: no sleep

src/mtev_http1.c
```diff
@@ -996,6 +1001,24 @@ mtev_http1_session_req_consume_read(mtev_http1_session_ctx *ctx,
...
+      if(head->allocd - head->start < MAX_CHUNK_HEADER_SIZE) {
+        mtevL(http_debug, "[fd=%d] Too little space in head node (%d). Reallocating.\n",
+              CTXFD(ctx), (int) (head->allocd - head->start));
+        struct bchain *new_head = ALLOC_BCHAIN(DEFAULT_BCHAINSIZE);
+        memcpy(new_head->buff, head->buff + head->start, head->size);
+        new_head->size = head->size;
+        new_head->start = 0;
+        new_head->compression = head->compression;
+        new_head->next = head->next;
+        /* It's safe to set {first,last}_input here, see (*@L1087) below. */
```
postwait: don't use a line number say "see the 'expand req.*_input list' comment below"

src/mtev_http1.c
```diff
@@ -996,6 +1001,24 @@ mtev_http1_session_req_consume_read(mtev_http1_session_ctx *ctx,
       }
     }
     else {
+      /* Check if we have enough in this node read the chunk header "$chunk_size\r\n" */
```
esproul: I don't understand this comment. Perhaps it should read "enough space in this node to read"?

src/mtev_http1.c
```diff
@@ -1052,6 +1082,17 @@ mtev_http1_session_req_consume_read(mtev_http1_session_ctx *ctx,
           ALLOC_BCHAIN(DEFAULT_BCHAINSIZE);
     }
     else if (tail->start + tail->size >= tail->allocd) {
+      /* expand req.*_input list
+       *
+       * (*) At least in the chunked case, we never get here.
+       * This code is executed only if the function is called with a full buffer before reading from
+       * the socket (below). However,
+       * - If we were holding a complete chunk in the buffer, we would have jumped this section (`goto` ~ L1038).
+       * - If we were holding an incomplete chunk, we expand the buffer to make it fit (`ALLOC_BCHAIN` ~ L1040).
```
postwait: Again, these line number references are fragile.  Use a more resilient reference method... C[1] C[2] is possible.

src/mtev_http1.c
HeinrichHartmann: uh. Indeed!

src/mtev_http1.c
HeinrichHartmann: fixed

src/mtev_http1.c
HeinrichHartmann: fixed

src/mtev_http1.c
```diff
@@ -996,6 +1000,24 @@ mtev_http1_session_req_consume_read(mtev_http1_session_ctx *ctx,
       }
     }
     else {
+      /* Check if we have enough space in this bchain node for the chunk header "$chunk_size\r\n" */
+      if(head->allocd - head->start < MAX_CHUNK_HEADER_SIZE) {
+        mtevL(http_debug, "[fd=%d] Too little space in head node (%d). Reallocating.\n",
+              CTXFD(ctx), (int) (head->allocd - head->start));
+        struct bchain *new_head = ALLOC_BCHAIN(DEFAULT_BCHAINSIZE);
```
postwait: I wonder if allocd is large enough, could we just memmove the start back to 0 in the same bchain without allocating a new one and freeing the old one.  The fact that we free the old one leads me to "yes".

HeinrichHartmann: I think that's true. Let me fix this.

HeinrichHartmann: Updated.

src/mtev_http1.c
```diff
@@ -996,6 +1000,15 @@ mtev_http1_session_req_consume_read(mtev_http1_session_ctx *ctx,
       }
     }
     else {
+      /* Check if we have enough space in this bchain node for the chunk header "$chunk_size\r\n" */
+      if(head->allocd - head->start < MAX_CHUNK_HEADER_SIZE) {
+        mtevL(http_debug, "[fd=%d] Too little space in head node (%d). Resetting bchain node.\n",
+              CTXFD(ctx), (int) (head->allocd - head->start));
+        memmove(head->buff, head->buff + head->start, head->size);
+        head->size = head->size;
```
postwait: noop

src/mtev_http1.c
```diff
@@ -996,6 +1000,15 @@ mtev_http1_session_req_consume_read(mtev_http1_session_ctx *ctx,
       }
     }
     else {
+      /* Check if we have enough space in this bchain node for the chunk header "$chunk_size\r\n" */
+      if(head->allocd - head->start < MAX_CHUNK_HEADER_SIZE) {
```
postwait: I think you still need the old code if this predicate fails.  And this is <= no?

src/mtev_http1.c
HeinrichHartmann: indeed...

src/mtev_http1.c
HeinrichHartmann: Regarding `<=`
- If we have equality then, the header should fit and we are good.
- In any case, MAX_CHUNK_HEADER_SIZE == 20, is just a guess, so `<` vs. `<=` is not an important edge case to consider.

If this condition fails, the current bchain node is large enough for our needs. The subsequent reads will be able to get the chunk header field into the buffer.
Once we have this, the section `if(str_in_f) { ... }` will prepare for the chunk payload.

HeinrichHartmann:
Expanding on the last comment. The whole loop could be redone, using a single large buffer that holds the latest chunk. Once we have completely read it, we move it to the user_data linked list (as we do now).
Looks like the {first,last}_input list is not doing much.

The reason I did not go for this refactor is, that I don't understand:
- (a) the non-chunked case
- (b) the reading of header fields, etc. in other http_request_* functions.

HeinrichHartmann:
Tests pass locally.

PR #561, opened 2019-08-12
Fix payload_complete when data is read but not yet consumed.

postwait:


HeinrichHartmann:
Still don't understand the how this fixes the issue you are seeing. Is this the call-site (mtev_http1.c:1721) ?

```
    while(ctx->conn.e && !mtev_http1_request_payload_complete(&ctx->req)) {
      int len = mtev_http1_session_req_consume(ctx, NULL, DEFAULT_BCHAINSIZE, 0, &mask);
      if(len < 0 && errno == EAGAIN) return mask | EVENTER_EXCEPTION;
      if(len <= 0) break;
    }
```

postwait:
that part wasn't broken, it is for consumers calling mtev_http_session_req_consume directly and attempting to determine completion.

HeinrichHartmann:
I'll give this a +1, since I understand the intention behind it, and the code is straight-forward.

Still not fully able to understand how we end-up in a situation where this would have worked before the last PR, and break after it. Don't want to spent too much time on this. If the tests pass, this is good to go.

PR #562, opened 2019-08-15
Fix implemenation leakage in lua_mtev.h header.

postwait:
The lua implementation leaked the presence of two debugging
variables into the header which meant that any consumers
of this header would have linkage issues that are legitimately
hard to fix correctly.


HeinrichHartmann:
Will need to fix a few other places, where we make use of those log streams. Update coming.

PR #564, opened 2019-08-22
support auto-updating consul kv tree attachment

postwait:


src/modules/consul.c
```diff
@@ -562,20 +716,24 @@ static void process_global_curlm(void) {
...
+              (const char *)mtev_dyn_buffer_data(&handler->dyn));
       }
  
       curl_multi_remove_handle(global_curl_handle, easy_handle);
       curl_easy_cleanup(easy_handle);
-      if(dyn) {
-        mtev_dyn_buffer_destroy(dyn);
-        free(dyn);
+      if(handler) {
+        mtev_dyn_buffer_destroy(&handler->dyn);
```
bdunavant: previously we would free "dyn" after mtev_dyn_buffer_destroy, but we no longer do.  Was this no longer necessary?

postwait: dyn is no longer allocated, it's now an incorporated member of the handler struct.

PR #568, opened 2019-08-27
Expose :tagged_name and :tagged_suppress for stats handles

postwait:


HeinrichHartmann:
LGTM.

I prefer using `luaL_checkstring(L, idx)` and `luaL_checkuserdata(L,idx, type)` instead of manual error checking. But this is a stylistic choice.

PR #572, opened 2019-08-30
Make SMR and ACO friendly.

postwait:
Given that ACO feels procedural but can span across many eventer loop
cycles, it is possible that an ACO consumer could quite unintentionally
create an SMR critical section (mtev_memory_{begin,end}) that spans
across the eventer loop boundary.  If this were to happen frequently,
SMR garbage collection could cease to make progress.  ck_epoch supports
this use case via ck_epoch_section_ts.  This change switches to use
ck_epoch_section_t for critical sections and attaches and transparently
activates a new section for each ACO coroutine.  We may want to lift the
mtev_memory_section_t and the swapping of these section into aco.c so
that aco could still be directly used outside the eventer.

### DIFF
--- a/src/eventer/eventer_aco_opset.c
+++ b/src/eventer/eventer_aco_opset.c
@@ -57,9 +57,16 @@ typedef struct aco_opset_info_t {
 } aco_opset_info_t;
 
 #define WORRISOME_STACK 2048
+static void
+eventer_aco_yield(void) {
+  aco_yield();
+}
 int
 eventer_aco_resume(aco_t *co) {
+  struct aco_cb_ctx *ctx = co->arg;
+  mtev_memory_section_t *prevs = mtev_memory_set_section(&ctx->section);
   aco_resume(co);
+  mtev_memory_set_section(prevs);
   if(co->save_stack.valid_sz > WORRISOME_STACK)
     mtevL(eventer_deb, "aco resume stack copy %zd bytes\n", co->save_stack.valid_sz);
   if(co->is_end) return eventer_aco_shutdown(co);
@@ -193,7 +200,7 @@ void
 eventer_aco_gate_wait(eventer_aco_gate_t gate) {
   bool zero;
   ck_pr_dec_32_zero(&gate->ref_cnt, &zero);
-  if(!zero) aco_yield();
+  if(!zero) eventer_aco_yield();
   free(gate);
 }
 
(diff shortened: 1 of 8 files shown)


src/eventer/eventer_aco_opset.c
```diff
@@ -57,9 +57,16 @@ typedef struct aco_opset_info_t {
 } aco_opset_info_t;
 
 #define WORRISOME_STACK 2048
+static void
```
cody271: Why are we wrapping `aco_yield()` in a new function here ?

src/utils/mtev_memory.c
```diff
@@ -69,7 +74,6 @@ static __thread int begin_end_depth = 0;
 static __thread uint64_t needs_maintenance = 0;
 static void *mtev_memory_gc(void *unused);
 static mtev_log_stream_t mem_debug = NULL;
-static pthread_mutex_t mem_debug_lock = PTHREAD_MUTEX_INITIALIZER;
```
cody271: Why was `mem_debug` lazy loaded with a lock before but not needed now ?

src/eventer/eventer_aco_opset.c
postwait: Makes it easier to debug.

src/utils/mtev_memory.c
postwait: Long ago, the log find didn't do an implicit create. Now, it is much more sane

src/eventer/eventer_jobq.c
```diff
@@ -330,7 +330,7 @@ eventer_jobq_create_internal(const char *queue_name, eventer_jobq_memory_safety_
 
 eventer_jobq_t *
 eventer_jobq_create_backq(const char *queue_name) {
-  return eventer_jobq_create_internal(queue_name, EVENTER_JOBQ_MS_CS, mtev_true);
+  return eventer_jobq_create_internal(queue_name, EVENTER_JOBQ_MS_GC, mtev_true);
```
cody271: What do these changes to the job queue do?

postwait: Requires the programmer to manage the critical sections.

It should help performance-wise.

cody271: Okay, that makes bit more sense now.

PR #577, opened 2019-09-15
Provide stacktrace hooks and implement a lua tracer

postwait:


HeinrichHartmann:
Nice! Let me see if I understand this correctly.

For each OS-thread we store a pointer to the currently active lua-coro in: `tls_active_lua_state`
mtev_stacktrace got a new hook that is able to replace/augment stack-frame information.
mtev_lua_stacktrace_assist is registered with this hook, to insert lua-stack information.

What's not clear to me is how the fame-level insertion works. I would have expected a single call to lua_taceback(L) that is appended to the very end of the stacktrace:
```
C-stack
...
lua-stack
...
```
- A single C-frame might be responsible for multiple lua-frames (lj_resume)
- There might be multiple C frames for a single lua-frame (e.g. object creation calls mutliple lj-functions)
Maybe this all works out. Just not clear to me how.

Also, by interleaving C-stack and lua-stack walking we might get in troubles when lua_getstack is called on a corrupted lua state: Would we still get a C-level stack trace if lua_getstack segfaults?

PR #579, opened 2019-09-18
Automatically attempt to ffi.cast pointers for hooks.

postwait:


src/modules/lua_general.c
```diff
@@ -762,6 +784,14 @@ mtev_lua_general_init(mtev_dso_generic_t *self) {
   }
   lua_settop(lmc->lua_state, 0);
 
+  lua_getglobal(lmc->lua_state, "require");
+  lua_pushstring(lmc->lua_state, "ffi");
+  if(lua_pcall(lmc->lua_state,1,1,0) != 0 || !lua_istable(lmc->lua_state,-1)) {
+    mtevL(mtev_error, "lua_general could reference ffi\n");
```
cody271:
```suggestion
    mtevL(mtev_error, "lua_general could not reference ffi\n");
```

src/modules/lua_general.c
```diff
@@ -762,6 +784,14 @@ mtev_lua_general_init(mtev_dso_generic_t *self) {
...
+    return -1;
+  }
+  lmc->ffi_index = luaL_ref(lmc->lua_state, LUA_REGISTRYINDEX);
```
cody271: `lmc->ffi_index` reference is never freed with `luaL_unref()`..

src/modules/lua_general.c
```diff
@@ -502,14 +502,35 @@ lua_hook_varargs(lua_State *L, const char *proto, va_list ap) {
...
-      //mtev_lua_setup_http_ctx(L, ctx);
     } else {
-      mtevL(nlerr, "unknown hook parameter type: '%s' consider mtev_lua_register_dynamic_ctype(...)\n", args[i]);
-      if(is_ptr) {
-        void *ptr = va_arg(ap, void *);
-        lua_pushlightuserdata(L,ptr);
+      bool is_ffi = false;
+      if(!is_ptr) {
+        /* we can't proceed b/c future arguments can't be popped off the varargs list correctly */
+        return i-1;
```
cody271: Should log an error here before returning.

src/modules/lua_general.c
postwait: This was intentional.. It lives until we close the lua state out for good.

src/modules/lua_general.c
postwait: hooks can fire fast! I made it a debug message so it's easier to silence.

PR #580, opened 2019-09-18
Support for OpenSSL 1.1

postwait:


src/eventer/eventer_SSL_fd_opset.c
```diff
@@ -1269,22 +1330,36 @@ int eventer_ssl_config(const char *key, const char *value) {
   }
   return 1;
 }
+
+static int32_t initialized = 0;
```
cody271: Needs `volatile` here? I think `eventer_ssl_init()` could be called from different threads..

src/modules/lua_mtev_crypto.c
```diff
@@ -84,7 +90,7 @@ mtev_lua_crypto_x509_index_func(lua_State *L) {
   cert = *((X509 **)udata);
   if(!strcmp(k, "signature_algorithm")) {
     int nid;
-    nid = OBJ_obj2nid(cert->sig_alg->algorithm);
+    nid = X509_get_signature_nid(cert);
```
cody271: What does this change here do?

src/modules/lua_mtev_crypto.c
```diff
@@ -576,7 +616,6 @@ static int __attribute__((unused)) mtev_lua_crypto_bn_##name(lua_State *L) { \
 static int mtev_lua_crypto_bn_is_negative(lua_State *L) {
   BN_METH_DECL_INT(0);
   lua_pushboolean(L, BN_is_negative(bn));
-  lua_pushboolean(L, bn->neg ? 1 : 0);
   return 2;
```
cody271:
```suggestion
   return 1;
```

src/utils/mtev_stacktrace.c
```diff
@@ -426,6 +426,11 @@ mtev_dwarf_walk_map(void (*f)(const char *, uintptr_t)) {
 #else
 #endif
 }
+#else
```
cody271: How are these changes related to OpenSSL ..?

src/eventer/eventer_SSL_fd_opset.c
postwait: It can't really.. We call this once, this is just protection in case a user is dumb and calls it again.

src/modules/lua_mtev_crypto.c
postwait: It doesn't change anything, but cert is opaque in openssl 1.1

src/utils/mtev_stacktrace.c
postwait: They aren't, I'll separate them. (same PR, but separate commit)

PR #581, opened 2019-09-19
Allow compilation without libdwarf

postwait:


src/utils/mtev_stacktrace.c
```diff
@@ -426,6 +426,11 @@ mtev_dwarf_walk_map(void (*f)(const char *, uintptr_t)) {
 #else
 #endif
 }
+#else
+const char *mtev_function_name(uintptr_t addr) {
+  (void)addr;
+  return NULL;
```
cody271: Should this return `NULL` or `""` ?

PR #585, opened 2019-09-25
Better coverity support.

postwait:
Add a simple though incomplete coverity model for libmtev consumers.
Change eventers calls that were #define maps into static inline
functions such that coverity models can apply to them.  Add a build
target for the coverity model file.

### DIFF
--- a/src/coverity_model.c
+++ b/src/coverity_model.c
@@ -1,0 +1,83 @@
+typedef void (*NoitHashFreeFunc)(void *);
+typedef int (*eventer_func_t)
+             (struct _event *e, int mask, void *closure, struct timeval *tv);
+
+int mtev_conf_get_string(void *base, char *path, char **val) {
+  int is_ok;
+  if(is_ok) {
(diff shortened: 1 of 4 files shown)


src/coverity_model.c
```diff
@@ -0,0 +1,83 @@
+typedef void (*NoitHashFreeFunc)(void *);
+typedef int (*eventer_func_t)
+             (struct _event *e, int mask, void *closure, struct timeval *tv);
+
+int mtev_conf_get_string(void *base, char *path, char **val) {
+  int is_ok;
+  if(is_ok) {
```
pamaddox: What is this doing? It looks like it's declaring an uninitialized variable, then immediately checking the value.

postwait: It's not C. it's just coverity modeling stuff.  It doesn't care what is_ok is... it is both true and false and the function is evaluated in the context of the return...  "what would happen in here if the function ultimately returns 1?"

PR #590, opened 2019-09-27
Prepare 1.8.2 release.

postwait:


ChangeLog.md
```diff
@@ -4,6 +4,17 @@
...
+### 1.8.2
+
+ * File-based (and stderr) logs now split newlines into separate
+   annotated log statements making logs easier to read.
+ * Address several crash issues with ACO+eventer interaction:
+   * "current" event is now ACO-local in addition to thread-local.
+   * http floating is bypassed for http connections in ACO mode.
+   * total removal of alloca() from the code base.
+ * The stock web ui javascript was fixed to prevent gratuitous API
+   usage when an instance is available after an interruption of service.
```
esproul: I don't quite understand what this means. What is an "instance" in this context, and what kind of interruption are we talking about?

postwait: When you using a web UI against an mtev app that leverages the mtev js stuff (snowth, fault, broker, etc.)...

if the instance goes down, but your browser stays open, it attempt to load various API endpoints every few seconds and they get QUEUED!  So when it is able to reconnect it will issue a flurry of backlogged requests for no good reason.

This fixes that.

esproul: Thanks, that makes complete sense.  How about this:

"The stock web ui javascript was fixed to prevent a flurry of queued API requests when the application comes back up after an interruption."

Not sure if that's any clearer, though.

PR #596, opened 2019-10-14
Add a http auth

postwait:


docs-md/apireference/c.md
```diff
@@ -4014,6 +4028,37 @@ mtev_hash_store(mtev_hash_table *h, const void *k, int klen, const void *data)
   mtev_hash_delete(), mtev_hash_delete_all() or mtev_hash_destroy().
  
 
+#### mtev_html_encode
+
+>Encode raw data as html encoded output into the provided buffer.
+
+```c
+int
+mtev_html_encode(const char *src, size_t src_len, char *dest, size_t dest_len)
```
bdunavant: If I send mtev_html_encode/decode empty strings (size 0), the destination is successfully encoded to an empty string (size 0), but our return code is now an error (0) even though the operation was successful.  Do we care about this case and need to return `-1` on an error?

docs-md/apireference/c.md
```diff
@@ -5270,6 +5315,70 @@ mtev_time_toggle_tsc(mtev_boolean enable)
...
+```
+
+
+  * `src` The buffer containing the encoded content.
+  * `src_len` The size (in bytes) of the encoded data.
+  * `dest` The destination buffer to which the function will produce.
+  * `dest_len` The size of the destination buffer.
+  * **RETURN** The size of the decoded output.  Returns zero is dest_len is too small.
+
+mtev_url_decode decodes input until an the entire input is consumed or until an invalid url-encoded character is encountered.
```
bdunavant: The current description does not explain what will happen in regards to the return code if this situation occurs.  Will I still get back 0 or a different return code?

docs-md/apireference/c.md
```diff
@@ -5270,6 +5315,70 @@ mtev_time_toggle_tsc(mtev_boolean enable)
...
+>Calculate how large a buffer must be to contain a decoded url-encoded string of a given length.
+
+```c
+size_t
+mtev_url_max_decode_len(size_t src_len)
+```
+
+
+  * `src_len` The size (in bytes) of the url-encoded string that might be decoded.
+  * **RETURN** The size of the buffer that would be needed to decode the input string.
```
bdunavant: What happens to mtev_url_max_decode_len if there is an invalid url-encoded character?

docs-md/config/modules/http_hmac_cookie.md
```diff
@@ -0,0 +1,57 @@
...
+   allowed: `/^.+$/`
+
+   A hex encoded key, if none is specified a random key will be
+   generated.
+
+ * **`max_age`** (optional)  [default: `86400`]
+
+   allowed: `/^\d+$/`
+
+   The number of seconds the cookie will be valid for (default 1
```
bdunavant: Is 0 valid here?

docs-md/config/modules/http_hmac_cookie.md
```diff
@@ -0,0 +1,57 @@
...
+   subdomain of the Host header if the Host header is at least three
+   units deep. (e.g. foo.com will not do anything, bar.foo.com will
+   set Domain=foo.com).
+
+ * **`user_(\S+)`** (option) 
+
+   allowed: `/^.*$/`
+
+   A user and password allowed.  This should be used for testing
+   only.
```
bdunavant: Possibly should use example here of how to add both.

docs-md/apireference/c.md
postwait: trying to keep this consistent across encode/decode (like b32 and b64)

docs-md/apireference/c.md
postwait: it doesn't take the input string, so it has a static response based on only the input size.

docs-md/config/modules/http_hmac_cookie.md
postwait: "valid" sure, it means auth won't work.

bdunavant: 0 in cookie life usually means "session" no?  Why wouldn't it work?

bdunavant: Quick googling.  0 should be delete, -1 is session. ok

PR #597, opened 2019-10-14
Console variants

postwait:


src/modules/lua_general.c
```diff
@@ -941,7 +943,8 @@ mtev_console_state_lua(mtev_console_closure_t ncct,
     return -1;
   }
   if(argc) {
-    eventer_set_owner(ncct->e, eventer_choose_owner(atoi(argv[0])));
+    if(ncct->simple)
+      eventer_set_owner(ncct->simple->e, eventer_choose_owner(atoi(argv[0])));
```
bdunavant: In honor of the recent linux `sudo` bug, do we trust `argv[0]` enough here that we don't need to check the return value?

src/mtev_rest.c
```diff
@@ -274,7 +274,7 @@ mtev_http_find_matching_route_rule(mtev_http_rest_closure_t *restc)
     int cnt;
     if (mtev_http_is_websocket(restc->http_ctx) == mtev_true) {
       if (strcmp(rule->method, "WS")) continue;
-      if (rule->websocket_protocol == NULL || protocol == NULL || strcmp(rule->websocket_protocol, protocol)) continue;
+      if (rule->websocket_protocol != NULL && protocol != NULL && strcmp(rule->websocket_protocol, protocol)) continue;
```
bdunavant: It looks like you were applying De Morgan's here but you did not negate the `strcmp` at the end.  Should there be a `!` in front of that, or was that in purpose?

postwait: it was in purpose.  This was a correction, so the new statement is different than the old.

PR #598, opened 2019-10-15
Fix logging zipkin with ACO and http2.

postwait:
* Add zipkin metadata to flatbuffer logs.
 * Fix zipkin traces if they move into ACO thread in mtev_rest.
 * Fix zipkin traces in http2.

### DIFF
--- a/src/mtev_http.c
+++ b/src/mtev_http.c
@@ -488,10 +488,12 @@ mtev_http_begin_span(mtev_http_session_ctx *ctx) {
   ctx->zipkin_span =
     mtev_zipkin_span_new(trace_id, parent_span_id, span_id,
                          mtev_http_request_uri_str(req), true, sampled, false);
+  mtevL(mtev_error, "BEGIN_SPAN(%p)\n", ctx->zipkin_span);
   mtev_http_connection *conn = mtev_http_session_connection(ctx);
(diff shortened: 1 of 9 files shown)


src/mtev_http.c
```diff
@@ -488,10 +488,12 @@ mtev_http_begin_span(mtev_http_session_ctx *ctx) {
   ctx->zipkin_span =
     mtev_zipkin_span_new(trace_id, parent_span_id, span_id,
                          mtev_http_request_uri_str(req), true, sampled, false);
+  mtevL(mtev_error, "BEGIN_SPAN(%p)\n", ctx->zipkin_span);
```
bdunavant: shoudl these be mtev_debug rather than error?

src/mtev_http.c
```diff
@@ -488,10 +488,12 @@ mtev_http_begin_span(mtev_http_session_ctx *ctx) {
...
   mtev_http_connection *conn = mtev_http_session_connection(ctx);
   if(conn) {
     eventer_t e = mtev_http_connection_event(conn);
     if(e) {
+  mtevL(mtev_error, "ATTACH_SPAN(%p -> %p)\n", ctx->zipkin_span, e);
```
bdunavant: same

PR #610, opened 2019-11-06
Add lua mtev.process:status

vasuraman11:
A process spawned by libmtev lua mtev.spawn may crash, and if the parent process is waiting for it to finish something lengthy, there was previously no function to just check status.  This PR adds a lua function mtev.process:status which allows a lua client to check the status so that it can be polled for a crash during a lengthy operation, especially useful in busted tests (right now the timeout is 5 minutes unless otherwise specified).

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -9,6 +9,7 @@
  * Allow crash stacktraces to be redirected optionally onto
    different log outlet
  * Add file_synch log type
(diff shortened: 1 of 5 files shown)


vasuraman11:
was able to avoid adding a new function by just using mtev.process:wait(0) and checking status to see if it is nil, and errno is also accessible, so this change is not needed

PR #616, opened 2019-11-15
Hoist the die tag test to avoid work.

postwait:


src/utils/mtev_stacktrace.c
```diff
@@ -283,8 +283,7 @@ static struct typenode *cache_type(struct dmap_node *node, Dwarf_Off off) {
   }
   return NULL;
 }
-static void extract_symbols(struct dmap_node *node, Dwarf_Die sib, mtev_log_stream_t dwarf_log) {
-  const char *tag_name;
+static void extract_symbols(struct dmap_node *node, Dwarf_Die sib, int depth, mtev_log_stream_t dwarf_log) {
```
vasuraman11: depth argument was added but is never used for anything, should it be part of a log print?

postwait: I used that in debugging.  It aligned the function signature better with the `register_die` function which has a depth (called `level`).  It is useful for maintainer debugging.

PR #628, opened 2019-12-10
MTEV_DIAGNOSE_CRASH can be set to an external tool path to be invoked with tid

vasuraman11:
In development and test, it can be convenient without requiring a parent to launch one of a number of debugging or reporting tools directly or via a script for mtev diagnose of a crash (instead of the internal stacktrace or backtrace coroner).  A small change allows this via MTEV_DIAGNOSE_CRASH directly.

### DIFF
--- a/src/utils/mtev_watchdog.c
+++ b/src/utils/mtev_watchdog.c
@@ -39,6 +39,8 @@
 #include <sys/ioctl.h>
 #include <fcntl.h>
 #include <sys/mman.h>
+#include <sys/types.h>
+#include <sys/syscall.h>
 #include <dirent.h>
 #include <execinfo.h>
 #if defined(__sun__)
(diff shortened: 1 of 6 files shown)


src/utils/mtev_watchdog.c
```diff
@@ -417,6 +419,24 @@ void mtev_self_diagnose(int sig, siginfo_t *si, void *uc) {
   raise(sig);
 }
 
+char *external_diagnose = NULL;
+void mtev_external_diagnose(int sig, siginfo_t *si, void *uc) {
+  (void)si;
+  (void)uc;
+  pid_t pid = syscall(SYS_gettid);
```
postwait: This looks like Linux only. There should be an alternative implementation that uses getpid()

vasuraman11: yes you are right, fixed.

PR #629, opened 2019-12-11
execlp needs null at end of args, add pid also when on linux, pid only non-linux

vasuraman11:


src/utils/mtev_watchdog.c
```diff
@@ -424,15 +424,19 @@ void mtev_external_diagnose(int sig, siginfo_t *si, void *uc) {
...
+  snprintf(tidstr, sizeof(tidstr), "%u", tid);
 #endif
   char pidstr[32];
-  snprintf(pidstr, sizeof(pidstr), "%u", pid);
+  snprintf(pidstr, sizeof(pidstr), "%u", getpid());
   int childpid = fork();
   if (!childpid) {
-    execlp(external_diagnose, external_diagnose, pidstr);
+#if defined(linux) || defined(__linux) || defined(__linux__)  
+    execlp(external_diagnose, external_diagnose, tidstr, pidstr, NULL);
```
postwait: I would reverse the arguments so that pid always the first argument and tid is optionally second

vasuraman11: done

PR #638, opened 2020-01-17
Make the lua mtev integration watchdog aware.

postwait:
Use thread-directed (Linux) POSIX timers to trigger a lua_hook
insertion into the state that stack traces and calls luaL_error().
This should prevent long-running lua states in a callback from
triggering the watchdog by breaking them.

### DIFF
--- a/src/modules/lua.c
+++ b/src/modules/lua.c
@@ -283,7 +283,6 @@ mtev_lua_resume(lua_State *L, int a, mtev_lua_resume_info_t *ri) {
   }
   if(ri && ri->lmc && !(ri->lmc->interrupt_time.tv_sec == 0 && ri->lmc->interrupt_time.tv_usec == 0)) {
     /* if we have an interrupt time specified for this lmc, then we should potentially reduce to it */
-    mtevL(mtev_error, "interrupt_time set...\n");
     if((diff.tv_sec == 0 && diff.tv_usec == 0) ||
         compare_timeval(ri->lmc->interrupt_time, diff) < 0) {
       diff = ri->lmc->interrupt_time;
--- a/src/modules/lua.c
+++ b/src/modules/lua.c
@@ -37,6 +37,9 @@
 #include <ck_pr.h>
 
 #include <unistd.h>
+#include <errno.h>
+#include <signal.h>
+#include <sys/syscall.h>
 
 #include "mtev_conf.h"
 #include "mtev_dso.h"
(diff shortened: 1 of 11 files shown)


src/modules/lua.c
```diff
@@ -191,19 +213,73 @@ mtev_lua_lmc_free(lua_module_closure_t *lmc) {
                      (const char*)&lmc->lua_state, sizeof(lmc->lua_state),
                      free, NULL);
     pthread_mutex_unlock(&mtev_lua_states_lock);
+    if(lmc->timer) {
+      timer_delete(lmc->_timer);
```
cody271:
```suggestion
  if (timer_delete(lmc->_timer) == -1) {
    mtevL(mtev_error, "timer_delete failed: %s", strerror(errno));
    return;
  }
```

HeinrichHartmann:
LGTM. If this works, this will be a big improvement for the lua situation.

The only comment I have, is that resetting the timeout in the mtev_lua_resume function will not catch all lua -> C yeidl operations. There are cases where we switch the lua state without going throught the eventer. We hit that with the nl_sleep_complete (where we now yield), if I remember correctly.
The current approach is fair, but this is something to keep in mind.

HeinrichHartmann:
Using `lua_sethook` to get into a consistent state after the signal handler was fired is VERY NICE. This was the piece I was missing.

With this technique, we should also be able to dump stack traces (without calling lua_error) on external signals. I would love to have a pstack tool for lua.

src/modules/lua.c
```diff
@@ -191,19 +212,88 @@ mtev_lua_lmc_free(lua_module_closure_t *lmc) {
...
+    if(diff.tv_sec < 0 || diff.tv_usec < 0) {
+      mtevL(nlerr, "lua attempting to resume with no time left before watchdog\n");
+      mtev_stacktrace(nlerr);
+      diff.tv_sec = 0;
+      diff.tv_usec = 1;
+    }
+  }
+  if(ri && ri->lmc && !(ri->lmc->interrupt_time.tv_sec == 0 && ri->lmc->interrupt_time.tv_usec == 0)) {
+    /* if we have an interrupt time specified for this lmc, then we should potentially reduce to it */
+    mtevL(mtev_error, "interrupt_time set...\n");
```
esproul: Should this perhaps be a debug log? I can't see what value it has in the normal error log.

HeinrichHartmann:
LGTM

PR #641, opened 2020-01-27
Address gcc9 warnings for a clean compile.

postwait:


src/utils/mtev_log.c
```diff
@@ -1047,8 +1047,8 @@ jlog_logio_cleanse(mtev_log_stream_t ls) {
     if(is_datafile(entry->d_name, &logid) &&  /* make sure it is a datafile */
        logid < earliest &&                    /* and that is older enough */
        logid != log->current_log) {           /* and that isn't current */
-      char fullfile[PATH_MAX];
-      char fullidx[PATH_MAX];
+      char fullfile[PATH_MAX+PATH_MAX];
+      char fullidx[PATH_MAX+PATH_MAX];
```
pamaddox: Why "PATH_MAX+PATH_MAX"?

postwait: making sure there's enough room to snprintf later.

src/noitedit/search.c
```diff
@@ -454,7 +454,7 @@ protected el_action_t
 cv_search(EditLine *el, int dir)
 {
 	char ch;
-	char tmpbuf[EL_BUFSIZ];
+	char tmpbuf[EL_BUFSIZ-3];
```
pamaddox: Any particular reason for dropping the size here?

postwait: to make sure it isn't too large to copy into the buffer it targets.

PR #647, opened 2020-03-04
reader-writer differentiation in mtev_conf

postwait:


src/mtev_conf.c
```diff
@@ -57,6 +64,10 @@
 #include "mtev_str.h"
 #include "mtev_thread.h"
 
+#if (__STDC_VERSION__ >= 201112L)
```
cody271: What does this do here?

src/mtev_conf.c
```diff
@@ -79,18 +90,111 @@ const char *_mtev_version = MTEV_VERSION;
...
 
-static pthread_mutex_t global_config_lock;
+/* This is tragically ugly.
+ *
+ * Ideally we would use a pthread_rwlock, however those are aware of only
+ * pthreads and we have aco threads to worry about as well.  We also need
+ * recursive reader locks so we use ck_rwlock_recursive_t locks here but
+ * they don't support recursive readers (despite having a call for that).
+ *
+ * The "work" but if you have a read lock and attempt to recursively acquire
```
cody271:
```suggestion
 * They "work" but if you have a read lock and attempt to recursively acquire
```

src/mtev_conf.c
```diff
@@ -79,18 +90,111 @@ const char *_mtev_version = MTEV_VERSION;
...
+ *
+ * In order to make the readlocks recursive we need a ACO-aware thread-local
+ * recursion counter for this lock and the whole global_recursion_{inc,dec}
+ * and _g(un)lock_read routines manage this.
+ *
+ * _glock is simpler b/c recursive write locks work fine.
+ *
+ * There still exist possible deadlock if the callers are not careful.
+ *
+ * If you old a read lock in an ACO thread and do something that yields
```
cody271:
```suggestion
 * If you hold a read lock in an ACO thread and do something that yields
```

src/mtev_thread.c
```diff
@@ -95,6 +96,12 @@ static mtev_boolean mtev_disable_binding = mtev_false;
 uint32_t
 mtev_thread_id(void) {
   if(!mtev_current_thread_id) mtev_current_thread_id = getthreadid();
+  aco_t *co = aco_co();
```
cody271: Should this use the new `aco_co_thread()` ?

postwait: yup

src/mtev_conf.c
postwait: _Static_assert isn't defined before c11.

PR #648, opened 2020-03-05
Implement shared waitfor/notify across threads

postwait:


src/modules/lua_mtev.c
```diff
@@ -5286,10 +5293,208 @@ mtev_lua_deserialize(lua_State *L, const lua_data_t *data){
...
+  }
+
+  /* There is no item yes, so we need to wait for one assuming timeout is non-zero */
+  if(timeout == 0) return 0;
+
+  mtev_lua_resume_info_t *ci;
+  ci = mtev_lua_find_resume_info(L, mtev_true); /* we cannot suspect without context */
+
+  struct shared_queues_wait *crutch = calloc(1, sizeof(*crutch));
+  double ntimeout = timeout < 0 ? 10900 : timeout;
```
cody271: Where does the value `10900` here come from?

src/modules/lua_mtev.c
```diff
@@ -5286,10 +5293,208 @@ mtev_lua_deserialize(lua_State *L, const lua_data_t *data){
...
+      mtev_lua_free_data((lua_data_t *)node->data);
+    }
+    else {
+      lua_pushnil(L);
+    }
+    free(node);
+    return 1;
+  }
+
+  /* There is no item yes, so we need to wait for one assuming timeout is non-zero */
```
cody271:
```suggestion
  /* There is no item yet, so we need to wait for one assuming timeout is non-zero */
```

src/modules/lua_mtev.c
```diff
@@ -5286,10 +5293,208 @@ mtev_lua_deserialize(lua_State *L, const lua_data_t *data){
...
+  if(timeout == 0) return 0;
+
+  mtev_lua_resume_info_t *ci;
+  ci = mtev_lua_find_resume_info(L, mtev_true); /* we cannot suspect without context */
```
cody271:
```suggestion
  ci = mtev_lua_find_resume_info(L, mtev_true); /* we cannot suspend without context */
```

src/modules/lua_mtev.c
postwait: Ha. Actually a typo, supposed to be 10000. But it is just an arbitrarily large timeout.

postwait: resolved in comment.

PR #656, opened 2020-04-09
Use condition variables instead of semaphores

postwait:


test/async-pingpong/test.c
```diff
@@ -33,8 +33,11 @@ usage(const char *prog) {
 static void
 parse_cli_args(int argc, char * const *argv) {
   int c;
-  while((c = getopt(argc, argv, "c:d")) != EOF) {
+  while((c = getopt(argc, argv, "c:di:")) != EOF) {
     switch(c) {
+      case 'i':
+        ITERS = atoi(optarg);
```
bdunavant, line 39: Should this have some error handling to avoid setting this to 0 if the user passes in garbage?

PR #663, opened 2020-04-29
Arch aarch64

postwait:


esproul:
Update: I can get it to build using our packaging script with these additional compiler flags, which quash errors stemming from libunwind's headers:
```
-Wno-error=unused-value -Wno-error=incompatible-pointer-types
```

postwait:
I did not build with libunwind.  Looking at the header, it would seem that error suppression is the only choice.

esproul:
Confirmed that I can build again without any special error suppression compiler flags.

PR #666, opened 2020-04-30
logic optimizations and string slices

postwait:


src/utils/mtev_logic.c
```diff
@@ -94,9 +94,11 @@ struct mtev_logic_var_t {
     int64_t     l;
     double      n;
   } value;
+  size_t s_len;
   pcre *re;
   pcre_extra *re_extra;
   mtev_boolean free_s;
+  char internal[40]; /* up 39 byte copies without alloc */
```
bdunavant, line 101: Wouldn't it be better to define 40 as a constant somewhere, and use it below instead of checking `sizeof(foo.internal)`?  Or does the compiler know that never changes and already replace it with a constant?

src/utils/mtev_logic.c
```diff
@@ -505,14 +543,19 @@ mtev_logic_exec_predicate(mtev_logic_exec_t *exec, mtev_logic_pred_t *pred, void
     }
   } else {
     if(lhs.type == MTEV_LOGIC_STRING) {
+      char ntcopy[128];
```
bdunavant, line 546: magic number from where?

src/utils/mtev_logic.c
postwait: Compiler knows. And sizeof is our coding style.

src/utils/mtev_logic.c
postwait: From me. I chose it because it seems big enough.

PR #669, opened 2020-05-01
Support ENV:default:{KEY} for config values

postwait:


src/mtev_conf.c
```diff
@@ -3776,8 +3776,39 @@ static void safe_free_xpath(void *in) {
...
+    char *key = strchr(fallback, ':');
+    if(key){
+      key++;
+      if(*key == '{' && key[strlen(key)-1] == '}') {
+        key++;
+        char key_copy[128];
+        snprintf(key_copy, sizeof(key_copy), "%.*s", (int)strlen(key)-1, key);
+        char *replace = getenv(key_copy);
+        if(replace) replace = strdup(replace);
+        else replace = strndup(fallback, key - fallback - 1);
```
vasuraman11, line 3798:
```suggestion
        else replace = strndup(fallback, key - fallback - 2);
```

vasuraman11: key was already on the colon and also incremented twice after, so you need to subtract 2 not 1

postwait: if the key was on the colon, we'd subtract 0... now that it has moved forward, we must subtract 1.

vasuraman11: it has moved forward twice

vasuraman11: 32322:{C
^            ^
|.            0x1007 (key)
0x1000 (fallback)
0x1007 - 0x1000 - 1 = 6
but the length of 32322 should be 5

postwait: You are correct... needed to remove the log as well

vasuraman11: ugh can't monospace font, the 0x1007 should be pointing to the first char of the env var, which is 2 past the colon

PR #672, opened 2020-05-06
CIRC-5070 Add backtrace/ptrace pretty printers

HeinrichHartmann:
This PR adds pretty-printers for the mtev_hash_table and mtev_http*_request structs in backtrace reports.
Here is an example how that looks in the backtrace-console:

* mtev_hash_table
image
* mtev_http_request attribute
image
* mtev_http_request thread annotation
image

To use these modules ptrace has to be invoked with `ptrace --modules-path=$(pwd)/backtrace ...`.

## TODO
- ~~Install those modules into an appropriate path (`/opt/circonus/share/backtrace/` ?)~~
- Change backwash/watchdog scripts to add `--modules-path` parameter


ChangeLog.md
```diff
@@ -4,6 +4,9 @@
 
 ## 1.10
 
+ * Add backtrace/ptrace pretty-printers for mtev_hash_table and
+   mtev_http*_request values
```
esproul, line 8: Should wrap the code names with backticks to prevent formatting weirdness due to the unbalanced '*'

HeinrichHartmann: indeed. _done_

HeinrichHartmann:
@esproul Do you have suggestions for the install locations?

HeinrichHartmann:
Also the location in the source tree (`/backtrace`) is up for debate.

esproul:
For the install location, how about `/opt/circonus/share/backtrace-plugins`?  I don't have an opinion about source repo location.

HeinrichHartmann:
Alright. I moved the files to `src/backtrace-support` since this is consistent with `src/mdb-support` which we already have and which serves a similar purpose.

As for the install location I would vote for `/opt/circonus/share/ptrace` since those are (backtrace-) ptrace modules. The `-plugins` / `-modules` postfix seems redundant, since I expect things in `../share/..` to be modules/libraries of some sort.

PR #673, opened 2020-05-07
 CIRC-5074 - Make mtev_cluster Config Update Synchronous

pamaddox:
When calling mtev_cluster_write_config, we were marking the config dirty
and returning success before actually writing out the config. This could
cause the config change to be lost if there's an ill-timed crash. Change
to force writing the config.


pamaddox:
Going to do this a different way

PR #674, opened 2020-05-07
CIRC-5074 - Add Flag To Cluster Endpoint To Force Conf Flush

pamaddox:
Add an optional flag, "x-mtev-conf-sync", that will allow forcing
cluster configuration changes to be written to disk immediately rather
than waiting for the scheduled write.

### DIFF
--- a/src/mtev_cluster.c
+++ b/src/mtev_cluster.c
@@ -1222,6 +1222,8 @@ rest_update_cluster(mtev_http_rest_closure_t *restc, int n, char **p) {
   int complete = 0, mask = 0, error_code = 500, status;
   xmlDocPtr indoc, doc;
   xmlNodePtr root;
+  mtev_hash_table *headers = NULL;
+  mtev_boolean force_conf_write = mtev_false;
 
   indoc = rest_get_xml_upload(restc, &mask, &complete);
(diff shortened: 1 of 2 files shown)


src/mtev_cluster.c
```diff
@@ -1222,17 +1222,39 @@ rest_update_cluster(mtev_http_rest_closure_t *restc, int n, char **p) {
...
   if(indoc == NULL) FAIL("xml parse error");
   root = xmlDocGetRootElement(indoc);
   if(!root || strcmp((const char *)root->name, "cluster"))
     FAIL("bad root node: not cluster");
+  headers = mtev_http_request_headers_table(mtev_http_session_request(ctx));
+  if (headers) {
+    char *force_conf_str = NULL;
+    char *hash_retr = "x-mtev-conf-sync";
+    if (mtev_hash_retrieve(headers, hash_retr, strlen(hash_retr), (void **)&force_conf_str)) {
+      int val = atoi(force_conf_str);
```
postwait, line 1239: This could be writte a bit clearer with the macros:
```
char *force_conf_str = mtev_hash_dict_get(headers, "x-mtev-conf-str");
if(force_conf_str) {
```

PR #680, opened 2020-05-13
Update autoconf tooling add coverage targets.

postwait:


esproul:
Same error on Ubuntu 16.04, with aclocal/automake 1.15.

PR #683, opened 2020-05-14
Fix support for NPN/ALPN negotiation on the server side.

postwait:
* Make http/1.1 an option for ALPN in the rest system
* Provide helpers to make it so we can register new alpn keys.
* Allow ssl contexts to override/add ALPN keys.
* More thorough HTTP tests

### DIFF
--- a/src/eventer/eventer_SSL_fd_opset.c
+++ b/src/eventer/eventer_SSL_fd_opset.c
@@ -99,6 +99,7 @@ struct eventer_ssl_ctx_t {
   unsigned no_more_negotiations:1;
   unsigned renegotiated:1;
   uint8_t *npn;
+  mtev_hash_table *alpn_funcs;
 };
 
 #define ssl_ctx ssl_ctx_cn->internal_ssl_ctx
@@ -602,6 +603,10 @@ eventer_ssl_ctx_free(eventer_ssl_ctx_t *ctx) {
   }
   if(ctx->npn) free(ctx->npn);
   if(ctx->ssl_ctx_cn) ssl_ctx_cache_node_free(ctx->ssl_ctx_cn);
+  if(ctx->alpn_funcs) {
+    mtev_hash_destroy(ctx->alpn_funcs, free, NULL);
+    free(ctx->alpn_funcs);
+  }
   if(ctx->issuer) free(ctx->issuer);
   if(ctx->subject) free(ctx->subject);
   if(ctx->cert_error) free(ctx->cert_error);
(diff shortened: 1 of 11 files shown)


src/eventer/eventer_SSL_fd_opset.c
```diff
@@ -937,39 +942,88 @@ static int next_proto_cb(SSL *ssl, const unsigned char **data,
...
-  int rv;
-  (void)ssl;
-  eventer_SSL_alpn_func_t f = (eventer_SSL_alpn_func_t)arg;
-
-  rv = f(out, outlen, in, inlen);
+  (void)arg;
+  if(!ssl) return SSL_TLSEXT_ERR_NOACK;
+  eventer_ssl_ctx_t *ctx = SSL_get_eventer_ssl_ctx(ssl);
+  if(!ctx) return SSL_TLSEXT_ERR_NOACK;
+  if(!ctx->npn) return SSL_TLSEXT_ERR_NOACK;
```
pjulien, line 949: Does this leak ``ctx``?

postwait: No, ctx is a refernce there and not allocated.

test/test_http_server.c
```diff
@@ -127,28 +127,90 @@ static int my_post_handler(mtev_http_rest_closure_t *restc, int npats, char **pa
...
+  } else {
+    cc = restc->call_closure;
+  }
+
+  int len = 0;
+  do {
+    int mask;
+    len = mtev_http_session_req_consume(ctx, cc->buffer, cc->buffer_size, cc->buffer_size, &mask);
+    if (len > 0) {
+      cc->seed = hash_djb2(cc->seed, cc->buffer, len);
```
pjulien, line 171: If there are multiple loops into this code, shouldn't the previously seen data be excluded?  Or is ``len`` here always the full size of the buffer?

i.e., if the data is read in two chunks of equal size, the first half of ``buffer`` would be passed in twice into this function

test/test_http_server.c
```diff
@@ -127,28 +127,90 @@ static int my_post_handler(mtev_http_rest_closure_t *restc, int npats, char **pa
...
+    }
+  } while(len > 0);
+  /* Avoid busy looping on the socket, spamming the debug logs */
+  usleep(cc->delay_us);
+  mtev_http_response_appendf(ctx, "HASH %" PRId64 "\r\n", cc->seed);
+  mtev_http_response_end(restc->http_ctx);
+  return 0;
```
pjulien, line 178: we're not releasing ``cc`` and ``cc->buffer`` because this is a test?

src/eventer/eventer_SSL_fd_opset.c
```diff
@@ -937,39 +942,88 @@ static int next_proto_cb(SSL *ssl, const unsigned char **data,
...
+  eventer_SSL_alpn_func_t f = NULL;
+  void *vf = NULL;
+  for (unsigned int i = 0; i < inlen; i += (unsigned int)(in[i] + 1)) {
+    if((ctx->alpn_funcs && mtev_hash_retrieve(ctx->alpn_funcs, in+i+1, in[i], &vf)) ||
+       mtev_hash_retrieve(&alpn_funcs, in+i+1, in[i], &vf)) {
+      f = vf;
+    }
+    if(f) {
+      int rv = f(out, outlen, in, inlen);
+      mtevL(ssldb, "alpn <- (%u:%u) %.*s -> %d\n", inlen, i, (int)in[i], in+i+1, rv);
```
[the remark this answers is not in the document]
postwait: %.*s is two args.  That's right.

src/eventer/eventer_SSL_fd_opset.c
```diff
@@ -937,39 +942,88 @@ static int next_proto_cb(SSL *ssl, const unsigned char **data,
...
+eventer_ssl_ctx_alpn_register(eventer_ssl_ctx_t *ctx, const char *name, eventer_SSL_alpn_func_t f) {
+  if(f == NULL) f = eventer_alpn_select_always;
+  if(!ctx->alpn_funcs) {
+    mtev_hash_table *t = calloc(1, sizeof(*t));
+    mtev_hash_init_locks(t, MTEV_HASH_DEFAULT_SIZE, MTEV_HASH_LOCK_MODE_MUTEX);
+    if(!ck_pr_cas_ptr(&ctx->alpn_funcs, NULL, t)) {
+      mtev_hash_destroy(t, free, NULL);
+      free(t);
+    }
+    t = ck_pr_load_ptr(&ctx->alpn_funcs);
```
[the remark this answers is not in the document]
postwait: You are correct.  Will fix.

test/test_http_server.c
postwait: always the size of buffer.

test/test_http_server.c
postwait: cc is attached with a free function to the restc that owns the request itself.  It will be freed when the context is completed.

PR #686, opened 2020-05-14
Various luamtev improvments.

postwait:
* Make luamtev print to stdout by default.
* Add -h formally as a no-error help.
* Make interactive mode work when not attached to a TTY.
* Fix -e flag for running different methods.
* Add tests.

### DIFF
--- a/src/luamtev.c
+++ b/src/luamtev.c
@@ -43,26 +43,28 @@ static char *droptouser = NULL, *droptogroup = NULL;
 static mtev_log_stream_t cli_stdout;
 
 static int
-usage(const char *prog) {
-  fprintf(stderr,
+usage(const char *prog, FILE *out) {
+  fprintf(out,
   "%s [-i] [-L luapath] [-C luacpath] [-M dir] [-d] [-e function]\n\tluafile\n\n", prog);
-  fprintf(stderr, "\t-u\t\t\tdrop to user\n");
-  fprintf(stderr, "\t-g\t\t\tdrop to group\n");
-  fprintf(stderr, "\t-i\t\t\tturn on interactive console\n");
-  fprintf(stderr, "\t-c <file>\t\tcustom mtev config\n");
-  fprintf(stderr, "\t-l <logname>\t\tenable the an mtev log stream\n");
-  fprintf(stderr, "\t-L <path>\t\tlua package.path\n");
-  fprintf(stderr, "\t-L +<path>\t\tappend to package.path\n");
-  fprintf(stderr, "\t-C <path>\t\tlua package.cpath\n");
-  fprintf(stderr, "\t-C +<path>\t\tappend to package.cpath\n");
-  fprintf(stderr, "\t-M <path>\t\tmtev modules path\n");
-  fprintf(stderr, "\t-m\t\t\tdaemonize and run managed\n");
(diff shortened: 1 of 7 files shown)


src/luamtev.c
```diff
@@ -139,11 +141,12 @@ make_config(void) {
...
 parse_cli_args(int argc, char * const *argv) {
   int c;
-  while((c = getopt(argc, argv, POSIXLY_COMPLIANT_PLUS "c:de:g:l:mn:iu:C:L:M:T")) != EOF) {
+  while((c = getopt(argc, argv, POSIXLY_COMPLIANT_PLUS "c:de:g:hl:mn:iu:C:L:M:T")) != EOF) {
     switch(c) {
       case 'd': debug = 1; break;
+      case 'h': usage(argv[0], stdout); exit(0); break;
       case 'i': interactive = 1; break;
-      case 'f': function = strdup(optarg); break;
+      case 'e': function = strdup(optarg); break;
```
HeinrichHartmann, line 149: whops. Good catch.

src/modules/lua_mtev.c
```diff
@@ -2776,6 +2777,20 @@ nl_print(lua_State *L) {
...
+*/
+static int
+nl_printto(lua_State *L) {
+  int n = lua_gettop(L);
+  if(n != 1) luaL_error(L, "mtev.printto(facility)");
+  char *tofree = print_tgt;
+  print_tgt = strdup(luaL_checkstring(L,1));
+  free(tofree);
+  return 0;
+}
```
HeinrichHartmann, line 2793: I wonder about the use-case of this. I have personally never used mtev.print(). Instead:
- print() -- from lua; or
- mtev.log()

Is this change in order to patch existing modules/apps?

src/luamtev.c
```diff
@@ -202,6 +208,13 @@ int luaopen_hostcli(lua_State *L) {
     lua_setfield(L, -2, key);
   }
   lua_setglobal(L, "ENV");
+
+  lua_getglobal(L, "mtev");
+  mtevAssert(lua_istable(L, -1));
+  lua_getfield(L, -1, "printto");
+  mtevAssert(lua_isfunction(L, -1));
+  lua_pushstring(L, "stdout");
+  lua_call(L, 1, 0);
```
HeinrichHartmann, line 217: So we change the default print target for luamtev apps here.

Is there a reason we don't globally change the print target for the whole mtev lua module?

src/modules/lua_mtev.c
postwait: libmtev's lua support replaces global `print` with `mtev.print` which defaults to "error".

This function just lets one (at an application level) change that default...

Since luamtev is run from the command line, it makes sense to use stdout.  When lua is running in a regular metv app, stdout makes little since.  But perhaps `error/lua/print` might make more sense as a default?  Something to think about later.

src/luamtev.c
postwait: This globally changes it.

HeinrichHartmann: right. But only for command line apps. The last comment explains the reasoning. OK.

src/modules/lua_mtev.c
HeinrichHartmann: that makes sense.

PR #688, opened 2020-05-14
accommodate misuse of env in mtev.exec/mtev.sh

postwait:


postwait:
got an out-of-band review from @HeinrichHartmann .

PR #690, opened 2020-05-17
SSL improvements.

postwait:
* Redo the tests to have intermediate certs.
* Improve ssl/crypto error messages.
* Support for inline key, cert, and ca in config.
* Separate ca_accept from ca_chain (for advertising)
* Extend the test suite to set SSL chained certs
* Disable compression in http/2 unless requested (to appease curl)
* Update lua bindings to accept a table and adapt HttpClient


docs/index.html
```diff
@@ -655,7 +655,7 @@ <h1>
                                 <section class="normal markdown-section">
                                 
                                 <h1 id=""><img src="assets/mtev-logo.png" alt="logo"></h1>
-<p>libmtev version 1.10.7</p>
+<p>libmtev version 1.10.2</p>
```
vasuraman11, line 658: why did this go from 7 to 2?

src/eventer/eventer_SSL_fd_opset.c
```diff
@@ -59,6 +60,64 @@
...
+    case SSL_ERROR_WANT_ASYNC: return "SSL_ERROR_WANT_ASYNC";
+#endif
+#ifdef SSL_ERROR_WANT_ASYNC_JOB
+    case SSL_ERROR_WANT_ASYNC_JOB: return "SSL_ERROR_WANT_ASYNC_JOB";
+#endif
+#ifdef SSL_ERROR_WANT_CLIENT_HELLO_CB
+    case SSL_ERROR_WANT_CLIENT_HELLO_CB: return "SSL_ERROR_WANT_CLIENT_HELLO_CB";
+#endif
+    default:
+      if(scratch) snprintf(scratch, len, "%d", err);
```
vasuraman11, line 115:
```suggestion
      if(scratch) snprintf(scratch, len, "unknown SSL_ERROR %d", err);
```

src/eventer/eventer_SSL_fd_opset.c
```diff
@@ -215,69 +282,68 @@ save_dh_params(DH *p, const char *filename) {
...
 
-static DH *dh1024_tmp = NULL, *dh2048_tmp = NULL;
-static const char *dh1024_file = NULL, *dh2048_file = NULL;
 static int
 generate_dh_params(eventer_t e, int mask, void *cl, struct timeval *now) {
   (void)e;
   (void)now;
-  int bits = (int)(intptr_t)cl;
+  struct stat sb;
+  char errstr[120];
```
vasuraman11, line 290: nit - magic 120 length like what Phil recently used although he set it at 65535

src/eventer/eventer_SSL_fd_opset.c
```diff
@@ -215,69 +282,68 @@ save_dh_params(DH *p, const char *filename) {
...
-    if(!dh2048_tmp) {
-      mtevL(mtev_notice, "Generating 2048 bit DH parameters.\n");
+    if(!tgt->params) {
+      mtevL(mtev_notice, "Generating %d bit DH parameters.\n", tgt->bits);
 #if OPENSSL_VERSION_NUMBER < 0x10100000L
-      dh2048_tmp = DH_generate_parameters(2048, 2, NULL, NULL);
+      tgt->params = DH_generate_parameters(tgt->bits, 2, NULL, NULL);
+      if(!tgt->params) {
+        while(0 != (err = ERR_get_error())) {
+          mtevL(eventer_err, " -> %s\n", ERR_error_string(err, errstr));
```
vasuraman11, line 312: here errstr is used, should also send the string length (120) so that it can be checked for overflow

src/eventer/eventer_SSL_fd_opset.c
```diff
@@ -703,7 +780,37 @@ eventer_ssl_ctx_t *
...
+  const char *ca = mtev_hash_dict_get(settings, "ca_chain");
+  if(!ca) mtev_hash_dict_get(settings, "ca_file");
+  const char *ciphers = mtev_hash_dict_get(settings, "ciphers");
+  const char *ca_accept = mtev_hash_dict_get(settings, "ca_accept");
+  if(!ca_accept) ca_accept = ca;
+  const char *crl = mtev_hash_dict_get(settings, "crl");
+  const char *npn = mtev_hash_dict_get(settings, "alpn");
+  if(!npn) npn = mtev_hash_dict_get(settings, "npn");
+  if(!npn) npn="h2,http/1.1";
+
```
vasuraman11, line 813: is there any logging to show what parameters mtev read in - that is important in order for a user to confirm that mtev took/parsed the correct settings into itself and avoid chasing your tail for hours

src/eventer/eventer_SSL_fd_opset.c
postwait: ERR_error_string doesn't take length, it expects the string to be long enough.  There is a newer API, but it would be worse (IMO) to use it b/c we'd need to ifdef to support older OpenSSLs.

src/eventer/eventer_SSL_fd_opset.c
postwait: 120 is specified by the OpenSSL docs and mentioned elsewhere in the code. I'll try to make this more obvious,.

src/eventer/eventer_SSL_fd_opset.c
postwait: In an effort to make the messages brief, this will make it say something like `ssl error: unkown SSL_ERROR %d`.. which isn't any better than `ssl error: %d`

docs/index.html
postwait: that'll get updated when we release

src/eventer/eventer_SSL_fd_opset.c
postwait: I'll add.

postwait: actually when set, it will print to ssldb deeper down.

src/eventer/eventer_SSL_fd_opset.c
vasuraman11: ah I didn't see the "ssl error:" prefix

docs/index.html
vasuraman11: ok

src/eventer/eventer_SSL_fd_opset.c
vasuraman11: that would be good, give it a #define

src/eventer/eventer_SSL_fd_opset.c
vasuraman11: ok no need, just #define the 120

src/eventer/eventer_SSL_fd_opset.c
vasuraman11: ok cool

PR #693, opened 2020-05-28
fix race in eventer startup resulting in unattended events

postwait:


src/eventer/eventer_impl.c
```diff
@@ -412,13 +412,21 @@ mtev_boolean eventer_watchdog_timeout_timeval(struct timeval *dur) {
   return mtev_watchdog_get_timeout_timeval(t->hb, dur);
 }
 
+static void *eventer_get_spec_for_thread(struct eventer_impl_data *t) {
+  if(t->spec == NULL) {
```
pjulien, line 416: This still races.  Need at the very least an acquire/release fence for getting and setting the value of the pointer.  This is the cheapest fence so any other stronger fence will also do.

pamaddox: We check the value again inside the lock - we see if null, acquire lock, then see if null to make sure it didn't get set in that brief window.

pjulien: You're incorrect, this still races.

pjulien: The pointer must be read and stored in a local variable using this [ck_pr_fence_acquire](http://concurrencykit.org/doc/ck_pr_fence_acquire.html)

Then it must be transferred from the local variable to the field using [ck_pr_fence_release](http://concurrencykit.org/doc/ck_pr_fence_release.html)

PR #697, opened 2020-06-02
add lua awareness of tls13

postwait:


src/modules/lua_mtev_crypto.c
```diff
@@ -125,31 +125,37 @@ mtev_lua_crypto_x509_index_func(lua_State *L) {
...
   if(!strcmp(k, "bits")) {
-#if OPENSSL_VERSION_NUMBER < 0x10100000L
     EVP_PKEY *pkey;
     pkey = X509_get_pubkey(cert);
     if (pkey == NULL) return 0;
-    else if (pkey->type == EVP_PKEY_RSA && pkey->pkey.rsa)
-      lua_pushinteger(L, BN_num_bits(pkey->pkey.rsa->n));
-    else if (pkey->type == EVP_PKEY_DSA && pkey->pkey.dsa)
-      lua_pushinteger(L, BN_num_bits(pkey->pkey.dsa->p));
+    else if (EVP_PKEY_base_id(pkey) == EVP_PKEY_RSA) {
```
esproul, line 131: Is this change because this more functional approach work all the way up through openssl 1.1.x?

postwait: The old ifdef section was wrong.  And this makes the legacy approach work on newer versions as well

PR #698, opened 2020-06-02
SSL verification was happening before the handshake was complete

postwait:


ChangeLog.md
```diff
@@ -4,6 +4,9 @@
 
 ## 1.11
 
+ * Fix an issues where valid SSL connections that do not handshake in one
+   non-blocking action would regsiter an spurious certificate error.
```
esproul, line 8:
```suggestion
 * Fix an issue where valid SSL connections that do not handshake in one
   non-blocking action would register a spurious certificate error.
```

PR #701, opened 2020-06-09
Cleaner JSON lib

postwait:


postwait:
This change fixes nad pointer returns in the mtev_json_util code and recorganizes the tree such that we only install the four necessary public headers.

ChangeLog.md
```diff
@@ -4,6 +4,12 @@
 
 ## 1.11
 
+### 1.11.4
+
+ * Fix more json-lib brokeness (bad pointers).
```
esproul, line 9:
```suggestion
 * Fix more json-lib brokenness (bad pointers).
```

PR #711, opened 2020-06-18
Move strlcpy and strlcat to utils.

postwait:
* Remove strlcpy and strlcat from noitedit/ to utils/
* Make them mtev_ prefixed
* If strlcat/strlcpy are not present on platform expose them weak


src/utils/mtev_str.h
```diff
@@ -79,6 +76,12 @@ API_EXPORT(void) mtev_str_buff_free(mtev_str_buff_t *buff);
 API_EXPORT(int) mtev_str_buff_len(mtev_str_buff_t *buff);
 API_EXPORT(char*) mtev_str_buff_to_string(mtev_str_buff_t **buff);
 
+#ifndef HAVE_STRLCPY
+API_EXPORT(size_t) strlcpy(char *dst, const char *src, size_t size);
```
pjulien, line 80:
```suggestion
API_EXPORT(size_t) strlcpy(char *restrict dst, const char *restrict src, size_t size);
```

etc.

src/utils/mtev_str.h
```diff
@@ -56,15 +56,12 @@ typedef struct mtev_str_buff{
 API_EXPORT(const char *) strnstrn(const char *, int, const char *, int) __attribute__((deprecated)); /*1.2.9*/
 #endif
 
-#if !defined(BROKEN_MEMMEM) && defined(_GNU_SOURCE)
-#define mtev_memmem(a,b,c,d) memmem(a,b,c,d)
-#else
 API_EXPORT(void *) mtev_memmem(const void *haystack, size_t haystacklen,
                                const void *needle, size_t needlelen);
-#endif
-
```
esproul, line 65: This seems unrelated?

postwait: Only slightly, but yes.  In the C file, there is the same logic where we call memmem.

I was trying to make sure we don't accidentally squat on what *should be* system stuff.

PR #713, opened 2020-06-18
Make __restrict work safely for C/C++ and use it.

postwait:


src/mtev_defines.h
```diff
@@ -89,6 +89,18 @@
 #  define SIZEOF_VOID_P SIZEOF_VOID__
 #endif
 
+/* Deal with the `restrict` keywork by making `__restrict` work */
```
esproul, line 92:
```suggestion
/* Deal with the `restrict` keyword by making `__restrict` work */
```

PR #714, opened 2020-06-19
Change MLKV{ } to MLKV( )

postwait:
This worked before mostly by accident because of the way CPP tokens
were pasted together.  Any subsequent array members were actually seen
as the varargs.  MLKV must take arguments to treat it as one parameter.
This changes the ABI and will break old compiles.  We now pass an array
of mtev_log_kv_t's to the logging functions instead of an array of
mtev_log_kv_t pointers.  Notably, this works within c++ now.

### DIFF
--- a/src/mtev_http.c
+++ b/src/mtev_http.c
@@ -627,7 +627,7 @@ mtev_http_log_request(mtev_http_session_ctx *ctx) {
     free(logline_dynamic);
   }
   mtevEL(http_access,
-        MLKV{ MLKV_STR("ip", ip), MLKV_STR("user", user ? user : "-"),
+        MLKV( MLKV_STR("ip", ip), MLKV_STR("user", user ? user : "-"),
               MLKV_STR("method", mtev_http_request_method_str(req)),
               MLKV_STR("protocol", mtev_http_request_protocol_str(req)),
               MLKV_STR("uri", mtev_http_request_uri_str(req)),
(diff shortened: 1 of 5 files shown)


src/mtev_http.c
```diff
@@ -636,7 +636,7 @@ mtev_http_log_request(mtev_http_session_ctx *ctx) {
               MLKV_DOUBLE("latency", (double)diff.tv_sec + (double)diff.tv_usec/1000000.0),
               MLKV_INT64("bytes_written", mtev_http_response_bytes_written(res)),
```
pjulien, line 637: looks like it should be `MLKV_UINT64`

src/mtev_http.c
```diff
@@ -636,7 +636,7 @@ mtev_http_log_request(mtev_http_session_ctx *ctx) {
...
               MLKV_INT64("bytes_read", mtev_http_request_content_length_read(req)),
```
pjulien, line 638: same here

src/mtev_main.c
```diff
@@ -603,19 +603,19 @@ mtev_main(const char *appname,
       pid_t owner;
       if((lockfd = mtev_lockfile_acquire_owner(lockfile, &owner)) < 0) {
         if(!wait_for_lock) {
-          mtevEL(mtev_stderr, MLKV{ MLKV_NUM("errno", errno), MLKV_END },
+          mtevEL(mtev_stderr, MLKV( MLKV_INT64("errno", errno), MLKV_END ),
                  "Failed to acquire lock: %s\n", lockfile);
           if(owner != -1) {
             pid_t groupid = getpgid(owner);
-            mtevEL(mtev_stderr, MLKV{ MLKV_NUM("pid", owner), MLKV_NUM("pgid", groupid), MLKV_END },
+            mtevEL(mtev_stderr, MLKV( MLKV_INT64("pid", owner), MLKV_INT64("pgid", groupid), MLKV_END ),
```
pjulien, line 610: I'm pretty sure these values are 32-bits, which is fine, but `pgid` I'm relatively sure is unsinged

PR #715, opened 2020-06-23
Various functional fixes for mtev_logic.

postwait:
* Set the string lenght on node creation.
 * Avoid allocation on tostring
 * Make strllcmp not return zero if lengths vary.

### DIFF
--- a/src/utils/mtev_logic.c
+++ b/src/utils/mtev_logic.c
@@ -111,34 +111,27 @@ static void mtev_logic_var_clean(void *vv) {
 }
 
 static inline const char *
-mtev_logic_var_tostring(mtev_logic_var_t *v, size_t *len) {
-  char buff[64];
-  if(v->type == MTEV_LOGIC_STRING) {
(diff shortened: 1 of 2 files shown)


src/utils/mtev_logic.c
```diff
@@ -320,10 +315,11 @@ static inline void *
 mtev_logic_ast_node_predicate(mtev_logic_ast_t *ast, const char *left,
                               mtev_logic_pred_op_t op, mtev_logic_var_t *right) {
   if(op == POP_RE || op == POP_NRE) {
+    char buff[4096];
```
bdunavant, line 318: Should this be like, max metric name length constant?

postwait: well, it doesn't have anything to do with a metric name, this is mtev and knows nothing about metrics.  It's just a sufficiently large buffer to use to stringify something... At some point we want to refuse to stringify b/c it would be too costly.

PR #724, opened 2020-07-24
-DMTEV_MEMORY_DEBUG will turn on debugging support

postwait:
Captures stack trace at alloc and free and then reports them if the
free looks invalid.

### DIFF
--- a/src/utils/mtev_memory.c
+++ b/src/utils/mtev_memory.c
@@ -43,6 +43,7 @@
 #include <umem.h>
 #endif
 #define MTEV_EPOCH_SAFE_MAGIC 0x5afe5afe
+#define MTEV_EPOCH_SAFE_FREE_MAGIC 0x00badbad
 
 #undef unlikely
 #define unlikely(x) (__builtin_expect(!!(x), 0))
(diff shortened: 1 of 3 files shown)


src/utils/mtev_memory.c
```diff
@@ -584,17 +592,34 @@ void *mtev_memory_ck_malloc(size_t r) {
...
 mtev_memory_ck_free_func(void *p, size_t b, bool r,
                          void (*f)(ck_epoch_entry_t *)) {
+  (void)b;
+  if(p == NULL) return;
   struct safe_epoch *e = (p - sizeof(struct safe_epoch));
 
-  if(p == NULL) return;
-  (void)b;
   bool magic_valid = e->magic == MTEV_EPOCH_SAFE_MAGIC;
+#ifdef MTEV_MEMORY_DEBUG
```
pjulien, line 600: Should be turned on by default when ASAN is detected

postwait: I don't think so.  This adds more memory: 128 per allocation.  I often run ASAN on bug setups where that can't be afforded.

I think it should be a runtime flag, but that will take more work

pjulien: all right

PR #726, opened 2020-07-24
this throws errors in clang8

postwait:


src/eventer/eventer_impl.c
```diff
@@ -415,16 +415,18 @@ mtev_boolean eventer_watchdog_timeout_timeval(struct timeval *dur) {
 }
 
 static void *eventer_get_spec_for_thread(struct eventer_impl_data *t) {
-  void *spec = atomic_load_explicit(&t->spec, memory_order_relaxed);
+  void *spec = t->spec;
+  if(spec != NULL) return spec;
```
pjulien, line 419: That doesn't work because of the previous load, give me a sec

postwait: I don't understand how it can not work.. if isn't NULL, it has to be correct. if it "is NULL, then we do the work to see if it is *REALLY* NULL and populate it like before.

pjulien: Because it's not CPU independent.  There's nothing in here that says reading is atomic, the value of the pointer can still tear.  We both know it doesn't on Intel but that's not the point.

PR #732, opened 2020-08-13
CIRC-5562 - Add Function To Drain Jobqs

pamaddox:
Add a new function, "eventer_jobq_drain_and_shutdown", that will wait
for all inflight/backlog jobs to finish, then close all of the jobq
consumer threads. This makes it safe to destroy the jobq.

### DIFF
--- a/src/eventer/eventer_jobq.c
+++ b/src/eventer/eventer_jobq.c
@@ -267,7 +267,7 @@ eventer_jobq_create_internal(const char *queue_name, eventer_jobq_memory_safety_
   jobq->mem_safety = mem_safety;
   jobq->isbackq = isbackq;
   jobq->lifo = &jobq_lifo_default;
-  jobq->shutdown_state = EVENTER_JOBQ_RUNNING;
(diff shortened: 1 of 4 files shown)


postwait:
The state, because it is being read and written from different threads needs to use ck_pr load and store.

src/eventer/eventer_jobq.c
```diff
@@ -267,6 +267,8 @@ eventer_jobq_create_internal(const char *queue_name, eventer_jobq_memory_safety_
   jobq->mem_safety = mem_safety;
   jobq->isbackq = isbackq;
   jobq->lifo = &jobq_lifo_default;
+  ck_pr_store_8(&jobq->shutdown_state, EVENTER_JOBQ_RUNNING);
```
postwait, line 270: This store doesn't really need to be atomic because no other threads have access to the jobq yet.  But it won't hurt antyhing.

pamaddox: I mostly just did this for consistency - make it clear we should always be doing this.

PR #735, opened 2020-09-18
Allow eventer config to be deep descendents

postwait:
This will allow
  <eventer>
    <intermediary>
      <config> ... </config>
    </intermediary>
  </eventer>


pamaddox (approved):
LGTM

PR #738, opened 2020-11-19
Quiet SSL layer not understood messages.

postwait:
If !sslv2 is specified and SSLv2 isn't available, then not being able
to turn it off isn't an error-worthy message.  If a "not understood"
SSL layer is encountered, use debug logging if negation is requested.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -4,6 +4,9 @@
 
 ## 1.12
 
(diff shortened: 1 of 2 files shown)


pamaddox (approved):

rizkybiz (commented):
👍

PR #740, opened 2020-11-25
DRAFT: CIRC-4436 - Allow stacktrace across an eventer asynch

vasuraman11:
This branch allows libmtev to stacktrace beyond `mtev_run_callback` and show what the stacktrace was when `eventer_add` was called.  That allows us to get a greater context if a crash occurs.

### DIFF
--- a/src/eventer/eventer.h
+++ b/src/eventer/eventer.h
@@ -49,6 +49,8 @@ extern "C" {
 #include <ck_pr.h>
 #include <ck_spinlock.h>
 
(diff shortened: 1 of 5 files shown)


postwait (commented):
This increases the eventer_t by over a kilobyte.

This should be implemented as an optional module that leverages the existing `eventer_context_t` as zipkin does.

postwait:
I also thing that zipking provides a lot of this already.  I think we should close this for a different approach.

PR #741, opened 2020-12-04
Respect `segment_size` and `precommit` jlog config

postwait:
The metadata alteration was done prior to the opening of the writer
which made it not work.  Delay the alteration until after the writer is
open.

CIRC-6004

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -2,12 +2,13 @@
 
 # 1
 
- * Fix missing top of crash stacktrace when libunwind is being used.
 
 ## 1.12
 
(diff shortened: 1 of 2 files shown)


esproul:
Curious if you think this will cause unexpected behavior; could an old modification suddenly "take effect" after this change, where it had been effectively dormant prior?

postwait:
We had intended long ago to change the segment size to 64MB in our of our apps, but found out that change was specified but never realized due to this.  So, yes, but that's the goal.

esproul (approved):
LGTM

PR #744, opened 2021-01-26
Ci lua hook timing

postwait:


pjulien:
LGTM

assuming Lua frees this automatically

pjulien (approved):

PR #747, opened 2021-02-01
Add console accessible stats to `mtev_fq`

postwait:


src/modules/mtev_fq.c
```diff
@@ -315,6 +337,73 @@ static logops_t fq_logio_ops = {
...
+  (void)argc;
+  (void)argv;
+  (void)dstate;
+  struct fq_module_config *conf = closure;
+  for(int i=0; i<conf->number_of_conns; i++) {
+    connection_configs *c = conf->configs[i];
+    fq_stats_t *s = &c->stats;
+    nc_printf(ncct, "== %s:%d ==\n  user: %s\n  exchange: %s\n  program: %s\n"
+                    "  (c) connects: %zu\n  (c) msgs rx: %zu\n  (c) msgs tx: %zu\n"
+                    "  (s) drops rx: %zu\n  (c) drops tx: %zu\n"
```
esproul, line 354: It's a small thing, but I'd like to see these two flipped in the ordering, as it will keep the client and server stats together.

With the current code lines 4-5 are "out of order":
```
  (c) connects: 1
  (c) msgs rx: 1062
  (c) msgs tx: 0
  (s) drops rx: 0
  (c) drops tx: 0
  (s) no exchange tx: 0
  (s) no routes tx: 0
  (s) routed tx: 0
  (s) drops tx: 0
  (s) msgs tx: 0
  (s) msgs rx: 1040
```

esproul (approved):
Small request, but won't hold up integration over it.

PR #748, opened 2021-02-08
Crash fix in lua_mtev cluster configuration

esproul:
Work by @postwait (Theo Schlossnagle <theo.schlossnagle@circonus.com>)


bdunavant (approved):
LGTM

PR #752, opened 2021-04-05
Allow configuration to continue without optional modules

esproul:
This does not change the disposition of the `--disable-{amqp,fq,zipkin-fq}` options, but instead makes their absence non-fatal. They are documented as optional modules, after all.  If the supporting library is present, the module will be built by default, just as it is today (i.e., no need to explicitly enable it).

Add a configuration summary message to make it clear whether any/all optional modules will be built.


bdunavant (approved):
LGTM

PR #753, opened 2021-04-06
CIRC-6436 - add yajl string encode

pjulien:


esproul (approved):
LGTM. Fixes the linker failure on Ubuntu.

PR #756, opened 2021-04-22
CIRC-6485 - C++ Compatible Ext Logging

pjulien:
- Introduce new compile time only template for logging array construction.
- Use C++ compatible aggregate construction
- This worked before mostly by accident because of the way CPP tokens
were pasted together.  Any subsequent array members were actually seen
as the varargs.  MLKV must take arguments to treat it as one parameter.
This changes the ABI and will break old compiles.  We now pass an array
of mtev_log_kv_t's to the logging functions instead of an array of
mtev_log_kv_t pointers.  Notably, this works within c++ now.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -5,6 +5,9 @@
 
 ## 1.12
 
+ * Update logging ext to work with C++.  This breaks the ABI
+   for existing users
+
 ### 1.12.18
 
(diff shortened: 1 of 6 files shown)


postwait (approved):
This is super. I like this a lot.

pamaddox (approved):

PR #757, opened 2021-05-13
CIRC-6566 - Fix race condition in mtev_memory_gc

pjulien:
This race condition makes it possible to access a pointer after it has been freed by another thread.  The problem is in `mtev_memory_gc`, the pointer `ar` is re-queued back to its original thread via:

```c
ck_fifo_spsc_enqueue(gc_return->queue, fifo_entry, ar);
```
but after this call, the current thread continues to use `ar`.  It is now a race between this thread finishing its use of the pointer and the other thread freeing it.

Sometimes the programs keeps on running with possible corruption, sometimes it crashes.  `ASAN` catches the issue, sometimes , but doesn't offer much help on the root cause unfortunately.

This crash affects `fault`, `snowth`, `stratcon` and the brokers.  There are literally hundreds of crashes spanning all these products in the last year because of this issue.  The [jira report](https://circonus.atlassian.net/browse/CIRC-6566) has sample reports

### DIFF
--- a/src/utils/mtev_memory.c
+++ b/src/utils/mtev_memory.c
@@ -364,14 +364,15 @@ mtev_memory_gc(void *unused) {
   const int max_setsize = 100;
   mtevL(mem_debug, "GC maintenance thread pid:%d starting.\n", getpid());
   while(ck_pr_load_int(&asynch_gc) == 1) {
-    struct asynch_reclaim *ar;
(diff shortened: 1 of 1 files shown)


postwait (approved):
Sweet.  Nice catch.

pamaddox (approved):
Great catch!

PR #760, opened 2021-06-03
CIRC-6386 - Race Condition

pjulien:
Fix race condition

https://circonus.atlassian.net/browse/CIRC-6294


src/eventer/eventer_jobq.c
```diff
@@ -109,6 +109,32 @@ static struct ck_malloc malloc_ck_hs = {
   .malloc = gen_malloc,
   .free = gen_free
 };
+static void
+eventer_jobq_consumer_notify(eventer_jobq_t *jobq) {
+  int64_t consumer_jobs;
+
+  pthread_mutex_lock(&jobq->consumer_lock);
+  jobq->consumer_jobs = MAX(jobq->consumer_jobs + 1, 1);
```
postwait, line 117: It seems all the MAX statement protect against a logic problem that should not happen.

pjulien: yes

PR #766, opened 2021-07-01
CIRC-6732 - Log Socket Pairs

pjulien:
Make it easy from "show reverse" to know which type of
sockets we're dealing with

https://circonus.atlassian.net/browse/CIRC-6732


vasuraman11 (approved):
LGTM

PR #776, opened 2021-09-27
allowed space(s) between number and unit string in duration parsing

andre9875:
[Jira ticket - CIRC-7085](https://circonus.atlassian.net/browse/CIRC-7085)

Currently, libmtev parsing function mtev_confstr_parse_duration can recognize only spaces between different time units like "1day 20min 10sec". New logic will allow end-users to use spaces freely everywhere like "1 day 20 min 10 sec".

### DIFF
--- a/src/utils/mtev_confstr.c
+++ b/src/utils/mtev_confstr.c
@@ -159,7 +159,6 @@ mtev_confstr_parse_duration(const char *input, uint64_t *output,
     /* white-space separated, null-terminated */
     while(*input && isspace(*input))
       input++;
-
     if(! *input)
       break;
 
--- a/src/utils/mtev_confstr.c
+++ b/src/utils/mtev_confstr.c
@@ -157,19 +157,8 @@ mtev_confstr_parse_duration(const char *input, uint64_t *output,
   *output = 0;
   while(true) {
     /* white-space separated, null-terminated */
-    /* sinse space(s) allowed between number and unit string */
-    /* the problem is if a user pass like "10 sec    " */
-    /* we are in infinitive loop according to the new logic line 169 - 170 */
-    /* assumption is to allow 5 times try before exit */
-    int count = 0;
-    while(*input && isspace(*input)) {
+    while(*input && isspace(*input))
       input++;
-      /* restore if space between number and unit string */
-      if(*input && !isdigit(*input) && count < 5) {
-        input--;
-        ++count;
-      }
-    }
 
     if(! *input)
       break;
--- a/src/utils/mtev_confstr.c
+++ b/src/utils/mtev_confstr.c
@@ -170,7 +170,8 @@ mtev_confstr_parse_duration(const char *input, uint64_t *output,
       return MTEV_CONFSTR_PARSE_ERR_FORMAT;
 
     /* remove space(s) between number and unit string */
-    while(!isalpha(*unit_str)){
+    while(*unit_str && !isalpha(*unit_str)){
+      if(isdigit(*unit_str)) return MTEV_CONFSTR_PARSE_ERR_FORMAT;
       unit_str++;
     }
 
--- a/src/utils/mtev_confstr.c
+++ b/src/utils/mtev_confstr.c
@@ -157,8 +157,20 @@ mtev_confstr_parse_duration(const char *input, uint64_t *output,
   *output = 0;
   while(true) {
     /* white-space separated, null-terminated */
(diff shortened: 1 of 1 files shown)


pjulien:
Why can't you just continue until you hit a non-space?

src/utils/mtev_confstr.c
```diff
@@ -157,8 +157,20 @@ mtev_confstr_parse_duration(const char *input, uint64_t *output,
   *output = 0;
   while(true) {
     /* white-space separated, null-terminated */
-    while(*input && isspace(*input))
+    /* sinse space(s) allowed between number and unit string */
+    /* the problem is if a user pass like "10 sec    " */
+    /* we are in infinitive loop according to the new logic line 169 - 170 */
```
pjulien, line 162: Not a good idea to put line numbers in comments since when the file gets edited the author will somehow have to remember to come here and edit the line numbers

src/utils/mtev_confstr.c
```diff
@@ -157,8 +157,20 @@ mtev_confstr_parse_duration(const char *input, uint64_t *output,
...
+    /* assumption is to allow 5 times try before exit */
+    int count = 0;
+    while(*input && isspace(*input)) {
       input++;
+      /* restore if space between number and unit string */
+      if(*input && !isdigit(*input) && count < 5) {
```
pjulien, line 168: don't get the '5'

src/utils/mtev_confstr.c
andre9875: I use line numbers only for this PR which will be removed after

src/utils/mtev_confstr.c
andre9875: it's irrelevant now. Find better way

src/utils/mtev_confstr.c
```diff
@@ -159,6 +159,7 @@ mtev_confstr_parse_duration(const char *input, uint64_t *output,
     /* white-space separated, null-terminated */
     while(*input && isspace(*input))
       input++;
+
```
pamaddox, line 162:
```suggestion
```

src/utils/mtev_confstr.c
```diff
@@ -169,6 +169,11 @@ mtev_confstr_parse_duration(const char *input, uint64_t *output,
     if(unit_str == input)
       return MTEV_CONFSTR_PARSE_ERR_FORMAT;
 
+    /* remove space(s) between number and unit string */
+    while(!isalpha(*unit_str)){
```
pamaddox, line 173: Is this guaranteed to not loop off the end of the string?

andre9875: add `while(*unit_str && !isalpha(*unit_str))`

src/utils/mtev_confstr.c
```diff
@@ -169,6 +169,12 @@ mtev_confstr_parse_duration(const char *input, uint64_t *output,
     if(unit_str == input)
       return MTEV_CONFSTR_PARSE_ERR_FORMAT;
 
+    /* remove space(s) between number and unit string */
+    while(*unit_str && !isalpha(*unit_str)){
```
pamaddox, line 173: There's no need to be using `isalpha` here. We don't care if a character is a letter or not; all we care about is if it is a space.

There's already a function in C, `isspace`, that will check exactly that. I'd recommend using that.

src/utils/mtev_confstr.c
```diff
@@ -169,6 +169,12 @@ mtev_confstr_parse_duration(const char *input, uint64_t *output,
...
+      if(isdigit(*unit_str)) return MTEV_CONFSTR_PARSE_ERR_FORMAT;
```
pamaddox, line 174: You don't need to check for a digit here if you just check for spaces in the initial loop. I'd recommend removing this line and handling all the error parsing after all the spaces have been skipped.

pamaddox (changes requested):
Requested changes are in individual comments.

src/utils/mtev_confstr.c
andre9875: switch to `isspace` function

src/utils/mtev_confstr.c
andre9875:
```    /* remove space(s) between number and unit string */
    while(*unit_str && isspace(*unit_str))
      unit_str++;```

pamaddox (approved):
LGTM

PR #785, opened 2021-12-28
CIRC-6131 Remove Alloca

mcieslinski:
grep says I got them all.

### DIFF
--- a/src/utils/android-demangle/cp-demangle.c
+++ b/src/utils/android-demangle/cp-demangle.c
@@ -4077,8 +4077,6 @@ cplus_demangle_print_callback (int options,
 
   d_print_flush (&dpi);
 
-
-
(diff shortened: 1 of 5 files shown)


src/utils/android-demangle/cp-demangle.c
```diff
@@ -4093,7 +4077,16 @@ cplus_demangle_print_callback (int options,
 
   d_print_flush (&dpi);
 
-  return ! d_print_saw_error (&dpi);
+
+
```
pamaddox, line 4081:
```suggestion
```

pamaddox: Should be a single blank line here, not three.

pamaddox, line 4081:
```suggestion
```

PR #794, opened 2022-02-01
Remove ex from all

postwait:


esproul:
Is the curl linkage commit supposed to be here?

esproul (approved):
LGTM

PR #797, opened 2022-02-07
Deprecated mtev atomic

bosley:
All public uses of volatile removed, and all items in mtev atomic marked as deprecated.

### DIFF
--- a/src/utils/mtev_atomic.h
+++ b/src/utils/mtev_atomic.h
@@ -40,18 +40,18 @@
 extern "C" {
 #endif
 
-//#warning "mtev_atomics is deprecated, use concurrencykit"
+#warning "mtev_atomics is deprecated, use concurrencykit"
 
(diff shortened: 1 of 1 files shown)


postwait:
If we're depreciating it, why are we changing it (remove of volatile) at the same time?

pjulien:
let's just remove the file and bump the mtev version

pjulien:
> If we're deprecating it, why are we changing it (remove of volatile) at the same time?

I think we should remove the file or fix the ``volatile`` but not both

postwait:
We're not trying to fix it, we're trying to get rid of it.  I support work on the latter, not the former.

pjulien:
> All public uses of volatile removed

It still has to work correctly, some of the removals are broken

postwait (commented):
please squash this.

PR #800, opened 2022-03-22
CIRC-7160 GCC 11 array-bounds warning

bosley:
This change fixes the array-bounds warning by allocating the entirety of the bchain structure. 

CI Has passed

### DIFF
--- a/src/mtev_http.c
+++ b/src/mtev_http.c
@@ -94,7 +94,7 @@ struct bchain *bchain_alloc(size_t size, int line) {
   struct bchain *n;
   /* mmap is greater than 1MB, inline otherwise */
   if (size >= 1048576) {
-    n = malloc(offsetof(struct bchain, _buff));
+    n = malloc(sizeof(struct bchain));
     if(!n) {
       mtevL(mtev_error, "failed to alloc bchain in bchain_alloc (size %zd)\n", size);
       return NULL;


src/mtev_http.c
```diff
@@ -94,7 +94,7 @@ struct bchain *bchain_alloc(size_t size, int line) {
   struct bchain *n;
   /* mmap is greater than 1MB, inline otherwise */
   if (size >= 1048576) {
-    n = malloc(offsetof(struct bchain, _buff));
+    n = malloc(sizeof(struct bchain));
```
mcieslinski, line 97: I would really love to understand what the intent of this was before and if the version of this that's in the `else` for this `if` should be changed too or if something else could/should be done.

pjulien: I doubt this is correct.  Most likely this is an alignment warning and the code went from being incorrect because of alignment to being just incorrect.  From now on please provide the warning you're getting

bosley: The warning type is posted in the title, and is listed in the ticket.

pjulien: That doesn't tell me anything, please give me the warning that the compiler printed out

bosley:
```
- compiling mtev_http.c
mtev_http.c: In function 'bchain_alloc':
mtev_http.c:102:6: error: array subscript 'struct bchain[0]' is partly outside array bounds of 'unsigned char[64]' [-Werror=array-bounds]
  102 |     n->type = BCHAIN_MMAP;
      |      ^~
mtev_http.c:97:9: note: referencing an object of size 64 allocated by 'malloc'
   97 |     n = malloc(offsetof(struct bchain, _buff));
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
mtev_http.c:103:6: error: array subscript 'struct bchain[0]' is partly outside array bounds of 'unsigned char[64]' [-Werror=array-bounds]
  103 |     n->buff = mmap(NULL, size, PROT_WRITE|PROT_READ, MAP_PRIVATE|MAP_ANON, -1, 0);
      |      ^~
mtev_http.c:97:9: note: referencing an object of size 64 allocated by 'malloc'
   97 |     n = malloc(offsetof(struct bchain, _buff));
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
mtev_http.c:109:6: error: array subscript 'struct bchain[0]' is partly outside array bounds of 'unsigned char[64]' [-Werror=array-bounds]
  109 |     n->mmap_size = size;
      |      ^~
mtev_http.c:97:9: note: referencing an object of size 64 allocated by 'malloc'
   97 |     n = malloc(offsetof(struct bchain, _buff));
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

pjulien: > I would really love to understand what the intent of this was before and if the version of this that's in the else for this if should be changed too or if something else could/should be done.

The idea is to not allocate the field ``_buf`` from ``struct bchain``.  If the buffer is too big, the content of the file are memory mapped and ``_buf`` is not allocated.

If the buffer is small enough, ``n`` is allocated with a size that can hold the buffer, by having ``_buf`` at the end it becomes the pointer to the buffer.  This allows the memory to stay together so can avoid a pointer jump when accessing ``n`` and then the data.

The code is correct as is but it's fine to waste the byte anyway because padding will make this struct 64-bytes total anyway

pjulien (approved):

PR #823, opened 2022-06-01
Make POST/PUT payload visible in backtrace and requests.json

vasuraman11:
The libmtev HTTP observer module and a backtrace Lua scripts have been changed to expose the incoming HTTP payload if it is text, so that we can see the incoming fetch REST API request for debugging Redfin.  This can also be useful in the case of other POST/PUT upload REST APIs, as long as the incoming payload is json.  Flatbuffer incoming payloads could be implemented in the future.


postwait:
Am I reading this right that we're going to pay a large allocation and memcpy on every request now? This seems like a bad default performance sacrifice.

Perhaps http end points should opt into this.

PR #848, opened 2022-08-30
CIRC-9069 - ub

pjulien:
Macros don't respect scope so it's not actually clear to me
what the expected behavior of

define CHANNEL cct->parent->data.channels[cct->channel_id]

since channel_id is a variable that can be freed and go out
of scope


vasuraman11 (approved):
LGTM good change

pamaddox (approved):

PR #859, opened 2022-09-30
Only Trigger On Reverse Socket Wakeup With Valid FD

pamaddox:


src/mtev_reverse_socket.c
```diff
@@ -620,8 +620,14 @@ mtev_reverse_socket_wakeup(eventer_t e, int mask, void *closure, struct timeval
...
   if (rc->data.e) {
-    eventer_remove_fde(rc->data.e);
-    eventer_trigger(rc->data.e, EVENTER_READ|EVENTER_WRITE);
+    int fd = eventer_get_fd(rc->data.e);
+    if (fd >= 0) {
+      eventer_remove_fde(rc->data.e);
+      eventer_trigger(rc->data.e, EVENTER_READ|EVENTER_WRITE);
+    }
+    else {
+      mtevL(nlerr, "%s: fd when attempting to wake up reverse socket was %d\n", __func__, fd);
```
esproul, line 629: Should this perhaps be a debug log, instead of error? I'm not sure what an operator would do with such information, and it sounds like it could happen frequently.

vasuraman11 (approved):
worried about potential log spam but it can be easily fixed if it does so LGTM

PR #860, opened 2022-09-30
Safer Usage Of eventer_remove_fd

pamaddox:
Check the fd in eventer_remove_fd and return NULL if it's less than zero. Add a check into reverse socket handling to check this.

### DIFF
--- a/src/eventer/eventer_epoll_impl.c
+++ b/src/eventer/eventer_epoll_impl.c
@@ -259,7 +259,7 @@ static void eventer_epoll_impl_update(eventer_t e, int mask) {
 static eventer_t eventer_epoll_impl_remove_fd(int fd) {
   eventer_t eiq = NULL;
   ev_lock_state_t lockstate;
-  if(master_fds[fd].e && fd >= 0) {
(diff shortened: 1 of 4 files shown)


esproul (approved):

src/eventer/eventer_epoll_impl.c
```diff
@@ -259,7 +259,7 @@ static void eventer_epoll_impl_update(eventer_t e, int mask) {
 static eventer_t eventer_epoll_impl_remove_fd(int fd) {
   eventer_t eiq = NULL;
   ev_lock_state_t lockstate;
-  if(master_fds[fd].e) {
+  if(master_fds[fd].e && fd >= 0) {
```
postwait, line 262: This uses a bad fd before checking it.

postwait (changes requested):

src/eventer/eventer_epoll_impl.c
pamaddox: Fixed

postwait (approved):

esproul (approved):
LGTM, sorry I missed that.

PR #874, opened 2022-10-26
Assert On Bad Refcnt

pamaddox:
Check to make sure the refcnt is above 0


pjulien (commented):
Just to be clear, if it's 0, that means the memory has been released, so there could be any garbage data in there, i.e., anything but zero

In other words, this doesn't do anything because it's freed when reading zero

pamaddox:
While that is technically true, I have seen numerous asserts in the codebase where the refcnt is zero. I am just trying to catch this earlier, despite the fact that there's no guarantee.

pjulien (approved):

PR #875, opened 2022-12-06
CIRC-9483 - heap use after free

pjulien:
The reverse socket (rc) is passed as a raw pointer via channel_closure_t.  This creates a race where
the rc can be shutdown and thus freed while an event is still pending delivery inside eventer

### DIFF
--- a/src/mtev_reverse_socket.c
+++ b/src/mtev_reverse_socket.c
@@ -560,7 +560,6 @@ mtev_reverse_socket_channel_handler(eventer_t e, int mask, void *closure,
     else if (parent_eventer && eventer_deref(parent_eventer)) {
       cct->parent->data.e = NULL;
     }
-
     free(cct);
     return 0;
   }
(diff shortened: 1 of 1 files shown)


src/mtev_reverse_socket.c
```diff
@@ -797,12 +799,15 @@ mtev_reverse_socket_handler(eventer_t e, int mask, void *closure,
         channel_closure_t *cct;
         cct = malloc(sizeof(*cct));
         cct->channel_id = rc->data.incoming_inflight.channel_id;
+        // Anchor 'rc' to the life cycle of 'newe'
+        mtev_reverse_socket_ref(&rc->refcnt);
```
mcieslinski, line 803: Am I missing something or is this meant to be passing `rc` rather than `&rc->refcnt`?

pjulien: yes, copy and paste error

src/mtev_reverse_socket.c
```diff
@@ -543,6 +545,8 @@ mtev_reverse_socket_channel_handler(eventer_t e, int mask, void *closure,
     else if (parent_eventer && eventer_deref(parent_eventer)) {
       cct->parent->data.e = NULL;
     }
+
+    mtev_reverse_socket_deref(&cct->parent);
```
mcieslinski, line 549: I think `parent` is already a pointer here

mcieslinski (approved):
Pretty cool

PR #890, opened 2023-01-11
CIRC-9644 - race

pjulien:
Pointer can be released after lock is dropped at any time because ref count was not bumped


mcieslinski (approved):
👍

postwait (approved):

pamaddox (approved):

PR #892, opened 2023-01-30
CIRC-9728 - Turn on "fanout" and "in_own_thread"

pjulien:
Make these options true by default for the http acceptor

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -4,7 +4,7 @@
 
 ## 2.4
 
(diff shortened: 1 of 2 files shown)


pamaddox (approved):

postwait (commented):
There is likely old code in (maybe noit) that relies on fanout being false as the default.

bdunavant (approved):
LGTM

postwait (approved):

PR #896, opened 2023-02-28
CIRC-9902 - align_watchdog_t

pjulien:
Align `mtev_watchdog_t` to a single cache line.  All monitored threads only have a single object that they can write to, it should not be possible in any scenario for one of these threads to stall out another thread via cache line interference

The only thread that can and should stall is the actual monitor thread since that's true sharing


vasuraman11 (approved):
LGTM, that is cool

PR #904, opened 2023-05-25
Fully adopt modern OpenSSL APIs

esproul:


test/speculatelog_test.c
```diff
@@ -1,8 +1,14 @@
 #include <mtev_defines.h>
 #include <mtev_log.h>
 #include <stdio.h>
+
+#define UNUSED(x) (void)(x)
```
RobBoeckermann, line 5: What's the reason for all of these "unused" functions? I'm assuming this is some weird compilation requirement?

test/subqueues/test.c
```diff
@@ -96,14 +101,17 @@ static int child_main(void)
...
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wcast-function-type"
   mtev_conf_watch_and_journal_watchdog((int (*)(void *)) mtev_conf_write_log, NULL);
+#pragma GCC diagnostic pop
 
   eventer_jobq_t *test_jobq = eventer_jobq_create_ms("test_jobq", EVENTER_JOBQ_MS_GC);
   eventer_jobq_set_concurrency(test_jobq, 4);
 
-  for(int i=0; i<sizeof(work)/sizeof(*work); i++) {
+  for(uint32_t i=0; i<sizeof(work)/sizeof(*work); i++) {
```
RobBoeckermann, line 112: general question: when do we use `uint32_t` over `size_t`?

RobBoeckermann (approved):

test/subqueues/test.c
esproul: For this change I was just trying to quash compiler warnings. In this case it was signed/unsigned mismatch and I was just trying to make the minimum change that worked. They could be changed to `size_t` if desired.

test/test_http_server.c
```diff
@@ -128,6 +130,9 @@ static int my_post_handler(mtev_http_rest_closure_t *restc, int npats, char **pa
   return EVENTER_READ | EVENTER_WRITE | EVENTER_EXCEPTION;
 }
 static int my_safe_handler(mtev_http_rest_closure_t *restc, int npats, char **pats) {
+  UNUSED(npats);
```
pjulien, line 133: I would rather adopt `C23` than add this everywhere.  Was there an error compiling this?

test/subqueues/test.c
pjulien: This should be ``size_t`` for these.  Should stick to the same integer size

test/pool-shift-async/test.c
```diff
@@ -97,7 +108,10 @@ child_main(void) {
   mtev_dso_post_init();
 
   mtev_conf_write_log();
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wcast-function-type"
```
pjulien, line 112: don't want pragmas like this in the code, please provide the error message, should be able to either add the appropriate cast (or remove it)

test/flowreg/test.c
```diff
@@ -130,7 +142,10 @@ static int child_main(void)
   mtev_dso_post_init();
 
   mtev_conf_write_log();
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wcast-function-type"
```
pjulien, line 146: please provide warning

test/async-pingpong/test.c
```diff
@@ -149,7 +164,10 @@ child_main(void) {
   mtev_dso_post_init();
 
   mtev_conf_write_log();
+#pragma GCC diagnostic push
```
pjulien, line 167: please provide warning

src/modules/lua_mtev_crypto.c
```diff
@@ -135,7 +138,23 @@ mtev_lua_crypto_x509_index_func(lua_State *L) {
...
+#if OPENSSL_VERSION_NUMBER >= _OPENSSL_VERSION_3_0_0
+    const int key_bits = EVP_PKEY_get_bits(pkey);
+    const int key_type = EVP_PKEY_get_base_id(pkey);
+
+    if (key_bits > 0) {
+      switch(key_type) {
+      case EVP_PKEY_RSA:
+      case EVP_PKEY_DSA:
+          lua_pushinteger(L, key_bits);
+          __attribute__((fallthrough));
```
pjulien, line 150:
```suggestion
```

don't want to use a GNU extension, can upgrade to ``C23`` though, and use ``[[fallthrough]]``

test/subqueues/test.c
esproul: OK will fix.

test/test_http_server.c
esproul: We can mark this as TODO once we have compiler support for C23. The issue was unused-variable warnings.

pjulien: I know, but in C23 you can do

```c
static int my_safe_handler(mtev_http_rest_closure_t *restc, int /*npats*/, char ** /*pats*/) {
```

or 

```c
static int my_safe_handler(mtev_http_rest_closure_t *restc, int , char **) {
```

or

```c
static int my_safe_handler(mtev_http_rest_closure_t *restc, [[maybe_unused]] int npats, [[maybe_unused]] char **pats) {
```

which matches C++.  I would rather do it the C way until then which is to say disable that warning for arguments than have this stuff everywhere instead of adding ``TODO`` everywhere

test/pool-shift-async/test.c
esproul: This is what happens without the pragma:
```
pool-shift-async/test.c: In function ‘child_main’:
pool-shift-async/test.c:111:40: error: cast between incompatible function types from ‘int (*)(void)’ to ‘int (*)(void *)’ [-Werror=cast-function-type]
  111 |   mtev_conf_watch_and_journal_watchdog((int (*)(void *))mtev_conf_write_log, NULL);
      |                                        ^
cc1: all warnings being treated as errors
make: *** [Makefile:60: pool-shift-async/test] Error 1
```

esproul: Offline we decided to create a new function with the correct signature and forwarding from there.

test/flowreg/test.c
esproul: This will be the same fix we settled on for the pool-shift-async test.

test/async-pingpong/test.c
esproul: This will be the same fix we settled on for the pool-shift-async test.

test/test_http_server.c
esproul: Decided to leave the unused-parameter warnings (not unused-variable, I misspoke above) but make them not errors. When we have C23 we can decorate them appropriately to quash the warning.

src/modules/lua_mtev_crypto.c
esproul: Addressed by adding an explicit `break` for each case.

test/flowreg/test.c
```diff
@@ -92,16 +92,12 @@ int test_job_cb(eventer_t e, int mask, void *v_work_description, struct timeval
...
-  if (mask == EVENTER_ASYNCH_WORK) {
-    work_description_t *work_description = (work_description_t *) v_work_description;
-  }
   return 0;
 }
 
 int test_poll(eventer_t e, int mask, void *v_work_description, struct timeval *now)
 {
   work_description_t *work_description = (work_description_t *) v_work_description;
-  unsigned int work_added = ck_pr_load_uint(&work_description->work_added);
```
pjulien, line 104: should be

```c
ck_pr_load_uint(&work_description->work_added);
```

to be equivalent

test/intern_test.c
```diff
@@ -175,7 +174,7 @@ int child_main() {
   for(int reps=0; reps<NREPS; reps++) {
   for(int p=0; p<2; p++) {
   pool = pools[p];
-  for(int wl=0; wl<WORKLOADS; wl++) {
+  for(uint32_t wl=0; wl<WORKLOADS; wl++) {
```
pjulien, line 177:
```suggestion
  for(size_t wl=0; wl<WORKLOADS; wl++) {
```

test/maybe_alloc_test.c
```diff
@@ -3,6 +3,7 @@
 #include <assert.h>
 #include <mtev_maybe_alloc.h>
 
+
```
pjulien, line 6:
```suggestion
```

test/flowreg/test.c
esproul: Done.

pjulien (approved):

PR #913, opened 2023-08-24
Fix Floated HTTP2 Sessions

pamaddox:
Code was resuming sessions that were explicitly marked asynch, resulting in firing too soon. Check to make sure the connection hasn't been floated before resuming.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -4,6 +4,8 @@
 
 ## 2.5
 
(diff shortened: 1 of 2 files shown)


postwait (changes requested):
Should include a changelog message

pamaddox:
> Should include a changelog message

Done

PR #917, opened 2024-02-13
Add Hooks For Reading/Writing Non-Standard Cluster Fields

pamaddox:
Add hooks to `mtev_clustering` to allow applications that use this to read and write custom fields in the cluster and node XML fields. This allows users to set things relevant to them, but are not necessarily relevant to all clusters.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -4,6 +4,8 @@
 
 ## 2.5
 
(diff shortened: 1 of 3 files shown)


bdunavant (approved):
LGTM

bdunavant (approved):
still lgtm

PR #919, opened 2024-02-28
Fix mtev_dyn_buffer_add_vprintf Bug

pamaddox:
There was a bug in `mtev_dyn_buffer_add_vprintf` where we could call `vsnprintf` with a `va_list` twice without a copy. This has undefined behavior, as the call to `vsnprintf` is destructive. Fixed it so that it does a copy now. However, that copy could have performance impacts, so also added a new function, `mtev_dyn_buffer_maybe_add_vprintf`, that will not try to increase the buffer size and requires the caller to handle it in order to avoid the extra copy.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -4,16 +4,24 @@
 
 ## 2.6
 
+ * Added new function, `mtev_dyn_buffer_maybe_add_vprintf`, that will not try to
+   expand the buffer and will return how many more bytes are needed.
+ * Fix bug in `mtev_dyn_buffer_add_vprintf` where the va_list argument could be used
+   after being modified. This change involves making a copy of the argument list
+   every time, so it could impact performance; recomment using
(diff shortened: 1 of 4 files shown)


ChangeLog.md
```diff
@@ -4,16 +4,24 @@
 
 ## 2.6
 
+ * Added new function, `mtev_dyn_buffer_maybe_add_vprintf`, that will not try to
+   expand the buffer and will return how many more bytes are needed.
+ * Fix bug in `mtev_dyn_buffer_add_vprintf` where the va_list argument could be used
+   after being modified. This change involves making a copy of the argument list
+   every time, so it could impact performance; recomment using
```
bdunavant, line 11:
```suggestion
   every time so it could impact performance; recommend using
```

bdunavant (approved):
LGTM

PR #922, opened 2024-02-29
Fix Compile Error On SSL Function Use For Some Compilers

pamaddox:
Some compilers complain that we're passing an uninitialized value to
openssl functions. These parameters are outparams, so it doesn't
actually matter, but initialize them to zero to make the compiler happy.


bdunavant (approved):
LGTM

PR #928, opened 2024-06-19
Include cstdint.h In mtev_json_object.h

pamaddox:
mtev_json_object.h uses things like uint64_t, which require ctdint.h. Not including it here makes other files have to include it and makes include order matter.

Added the include directly to mtev_json_object.h to avoid this.


esproul (approved):
LGTM

PR #930, opened 2024-08-20
Crypto/LibSSL Usage Fixes

pamaddox:
* Check for errors when calling X509_get1_ocsp and handle them
* Fix bug where we were reporting errors on post-2038 times even when 64-bit time is available.


bdunavant (approved):
LGTM

PR #933, opened 2025-01-31
Kafka Support/Module

pamaddox:
Added a kafka module that will allow registering a hook and reading from a configured kafka topic.

Support for writing data out is forthcoming.


src/modules/mtev_kafka.cpp
```diff
@@ -0,0 +1,428 @@
...
+  }
+  the_conf = new kafka_module_config{};
+  mtev_image_set_userdata(&self->hdr, the_conf);
+  return the_conf;
+}
+
+static int kafka_logio_open(mtev_log_stream_t ls)
+{
+  (void) ls;
+  return 0;
```
postwait, line 341: This should return -1 until it is implemented.

src/modules/mtev_kafka.cpp
```diff
@@ -0,0 +1,428 @@
...
+}
+
+static int
+  kafka_logio_write(mtev_log_stream_t ls, const struct timeval *whence, const void *buf, size_t len)
+{
+  // TODO: Need a write function
+  return len;
```
postwait, line 348: This should return -1 until an implementation exists.

postwait (commented):
Looks fairly simple.

Asside from the logops comments, I would suggest collpasing the changeset to a single commit.  Fishing through 91 commits can be painful.

src/modules/mtev_kafka.cpp
pamaddox: Done

src/modules/mtev_kafka.cpp
pamaddox: Done

PR #936, opened 2025-03-07
Add Extra Fields To Kafka Processing Config

pamaddox:
Allow setting arbitrary fields in the config so users can set fields specific to their applications. Also added a "protocol" field for configuring the expected format of incoming data.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -4,6 +4,9 @@
 
 ## 2.7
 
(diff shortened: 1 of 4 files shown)


postwait (changes requested):
I think this should all be documented in the mtev_kafka.xml docs.

postwait (approved):
LGTM

PR #943, opened 2025-03-31
Kafka Module Example Programs

pamaddox:
Add a kafka_producer and kafka_consumer test program that allows setting up both a consumer and producer and using them to send and receive messages.

Added topic to consumer hook to make it clear where messages came from.

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -4,6 +4,9 @@
 
 ## 2.7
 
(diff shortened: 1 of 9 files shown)


postwait (changes requested):
This would appear to make rdkafka a hard requirement for libmtex now as it is an unconditional all: target in the Makefile.

postwait (approved):

PR #945, opened 2025-04-04
Switch Timezone Test To Use America/New_York

pamaddox:
US/Eastern is a symlink to America/New_York and is no longer support in some later tzinfo packages. Switch the test to explicitly check for America/New_York, which is.


esproul (approved):
LGTM, thanks!

PR #947, opened 2025-06-11
 ci-12789.mtevfatal _log_line_number_and_filename

rameshbabu892:


pamaddox (changes requested):
This code does not compile. If you attempt to compile it, you get numerous errors stating that `mtevFatal` is incorrect:

```
../src/utils/mtev_log.h:854:15: error: format ‘%d’ expects argument of type ‘int’, but argument 6 has type ‘const char *’ [-Werror=format=]
  854 |   mtevL((ls), "[FATAL][" __FILE__ ":%d] " args, __LINE__);\
      |               ^~~~~~~~~~
```

You shouldn't need to change any call sites; you should be able to change `mtevFatal` so that a line number and file are provided.

